PR #3, opened 2021-01-16
fix: mysql documentation

mfts:
Hi Jake, 

I love the new docs, especially love the use of Next.js. 

I saw that MySQL docs hasn't been updated. I don't remember exactly what variables you are using for MySQL so I added the ones I thought make sense based on Mongo, PG, Redis docs.

✌️ Marc


coffee-cup (approved):
Thanks for the PR Marc!

PR #8, opened 2021-02-10
fix typo is redis.md

dannylindquist:


coffee-cup (approved):
Thanks for the fix! 😄

PR #21, opened 2021-03-12
Update cli docs to reflect reality (disconnect/connect commands)

queer:


src/pages/cli/api-reference.md
```diff
@@ -20,11 +20,12 @@ new project, or select an existing project.
 Run a command inside of the Railway environment.
 
 ```bash
-railway run <cmd>
+railway run [cmd]
```
coffee-cup, line 23: I think we have `railway run <cmd>` in a few other places as well. Can you please change them as well :)

coffee-cup (approved):

PR #30, opened 2021-04-28
Document the railway environment variables

gschier:
This PR adds a new Deployments > Environment Variables doc to explain the environment variables that Railway provides to builds/deploys.

This PR _also_ changes the default port from `3000` to `3001` to prevent conflicting with other dev services.

image

### DIFF
--- a/src/pages/deployment/environment-variables.md
+++ b/src/pages/deployment/environment-variables.md
@@ -7,7 +7,6 @@ builds and deployments.
 
 | Name                         | Description |
 |------------------------------|-------------|
-| `RAILWAY_DEPLOYMENT_URL`     | The domain specific to the current deployment, of the form `c56c8f2a83d2.up.railway.app` (not available during builds) |
 | `RAILWAY_STATIC_URL`         | The public domain, of the form `example.up.railway.app` |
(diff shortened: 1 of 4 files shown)


src/pages/deployment/environment-variables.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Environment Variables
+---
+
+Railway provides the following additional system environment variables to all 
+builds and deployments.
+
+| Name                         | Description |
+|------------------------------|-------------|
+| `RAILWAY_DEPLOYMENT_DOMAIN`  | The domain specific to the current deployment, of the form `c56c8f2a83d2.up.railway.app` (not available during builds) |
```
JakeCooper, line 10: So, if we switch to deploymentID this could be available during builds

JakeCooper: But I suppose we can update this later when it's actually avail

JakeCooper (commented):
I kinda like PORT 3k :S

src/pages/deployment/environment-variables.md
gschier: Turns out it wasn't even available to the deploy because it's generated after. Removing for now until after the domains refactor.

PR #41, opened 2021-06-11
update docs for environment variable description

GitBolt:


gschier (approved):
Ooo, nice catch @GitBolt, and thanks for submitting it 👍🏻 🥳

PR #48, opened 2021-07-29
Version info for Node and Python

coffee-cup:
Docs taken from

https://devcenter.heroku.com/articles/python-runtimes

and

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines


JakeCooper (approved):
Oh hell ya

PR #49, opened 2021-08-03
(1204): Clarified Python Project Procfile Requirements

ndneighbor:
_Say that 5 times, 5 times as fast_

Thanks to Discord user: andreawthaderp#3031 for catching this issue. Added copy to clarify the Python Procfile requirements.

Comments from the Author
- Might be good to call out that we look for `web:` in the file but I will spend some time thinking about the bigger picture of docs before including that.

### DIFF
--- a/src/pages/deployment/builds.md
+++ b/src/pages/deployment/builds.md
@@ -48,7 +48,11 @@ The [Python buildpack](https://github.com/heroku/heroku-buildpack-python)
 detects if your build is Python by looking for a `requirements.txt` file. If
 found, dependencies will be installed using `pip`.
 
-Please include a [Procfile](/deployment/builds#procfile) in the root folder of your repository. If no [Procfile](/deployment/builds#procfile) is found, your deploy might fail to start. 
+Please include a [Procfile](/deployment/builds#procfile) in the root folder of your repository. If no [Procfile](/deployment/builds#procfile) is found, your deploy might fail to start with the following error.
+
+```
+ERROR: failed to launch: determine start command: when there is no default process a command is required
+```
 
 The default Python version is `3.6`.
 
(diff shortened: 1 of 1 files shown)


src/pages/deployment/builds.md
```diff
@@ -48,6 +48,8 @@ The [Python buildpack](https://github.com/heroku/heroku-buildpack-python)
 detects if your build is Python by looking for a `requirements.txt` file. If
 found, dependencies will be installed using `pip`.
 
+Please include a [Procfile](/deployment/builds#procfile) in the root folder of your repository. If no [Procfile](/deployment/builds#procfile) is found, your deploy might fail to start.
```
JakeCooper, line 51: Might want to replace it with the actual error

That way when people are searching for that specific error, whether docs or SEO stuff, it'll come up

JakeCooper (approved):

src/pages/deployment/builds.md
ndneighbor: Can you check now to see if it's preferable?

src/pages/deployment/builds.md
```diff
@@ -92,8 +98,7 @@ process1: command
 process2: command
```
coffee-cup, line 98: Might want to remove this bit too since we don't support multiple processes

ndneighbor: Lemme fix that

src/pages/deployment/builds.md
```diff
@@ -94,11 +94,10 @@ the deployment starts.
 A Procfile is in the format of
 
 ```
-process1: command
-process2: command
+process: command
```
coffee-cup, line 97: very minor nit: Use `web` here since we should push users in that direction since it works for web servers

coffee-cup (approved):

src/pages/deployment/builds.md
JakeCooper: What do you mean?

src/pages/deployment/builds.md
ndneighbor: Unsure if we should do that as it's explaining a procfile, also, the web Procfile sample is directly after

PR #56, opened 2021-08-23
Domain Addition and QoL fixes

ndneighbor:
**Added**

- Google Domain Caveat
- PORT information under the starter
- FAQ updates to support the ability to render links


src/pages/getting-started.md
```diff
@@ -57,3 +56,11 @@ environment. Click the returned link to see the build and deploy logs.
...
 You can also setup [auto deploys](/deployment/github-triggers) so that a deploy
 is created everytime you push to a branch.
+
+### Exposing Your Application
+
+Before your application can say hello, Railway needs to know what PORT to listen on to expose your application to the internet. Railway does try to do it's best to do this automatically for you however there are cases when we can't.
+
+You can configure your application to use the `PORT` environment variable. You can add the `PORT` variable under your project's variables. (Command + K and type `Variables` or you can use the keyboard shortcut: `G` + `V` under your selected project)
+
+Keep in mind, when you publish your app, make sure that your application's IP is not set to `127.0.0.0`, Railway won't be able to expose your application if it's set to that IP.
```
JakeCooper, line 66: We should mention what users SHOULD do not what they shouldn't do

ndneighbor: Just to confirm: the recommendation is that users should be using 0.0.0.0 as the IP?

PR #66, opened 2021-09-11
Feedback and layout fixes

ndneighbor:
We now have a direct feed of feedback. I would also like to give a shoutout to CORS.


src/components/Icon.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
```
coffee-cup, line 1: Not sure if this component is needed since we just import the icons from react feather

src/layouts/DocsLayout.tsx
```diff
@@ -27,6 +33,26 @@ export const DocsLayout: React.FC<Props> = ({
       `https://github.com/railwayapp/docs/edit/main/src/pages${pathname}.md`,
     [pathname],
   );
+  const [helpful, setHelpful] = useState(true);
+  const [success, setSuccess] = useState(false);
+
+  const sendFeedback = async event => {
+    event.preventDefault()
+    const feedback = event.target.feedback ? event.target.feedback.value : "";
```
coffee-cup, line 41: nit: Could update to
```suggestion
    const feedback = event.target.feedback ?? "";
```

coffee-cup (approved):
Looks good!

src/components/Icon.tsx
ndneighbor: Banner needs it, will remove on refactor

PR #68, opened 2021-09-15
Save feedback to postgres

coffee-cup:
- Setup prisma
- Save feedback responses to Postgres
- Only send negative resopnses to Discord
- Split feedback UI into separate component
- Fix z-index of model


src/pages/api/feedback.ts
```diff
@@ -0,0 +1,50 @@
...
+  feedback?: string;
+}
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
+  if (req.method == "POST") {
+    const body: FeedbackBody = JSON.parse(req.body);
+    const feedbackMessage =
+      body.feedback?.trim() === "" ? null : body.feedback?.trim();
+
+    const discordBody = {
```
FarazPatankar, line 17: Maybe create this only if we're sending the message to Discord?

FarazPatankar (approved):
LGTM

prisma/migrations/20210915164107_feedback/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- CreateEnum
+CREATE TYPE "Sentiment" AS ENUM ('POSITIVE', 'NEGATIVE');
+
+-- CreateTable
+CREATE TABLE "Feedback" (
+    "id" TEXT NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
ndneighbor, line 8: For my knowledge, why would there be an updated at if there is no editing on these comments

ndneighbor (approved):

prisma/migrations/20210915164107_feedback/migration.sql
coffee-cup: It probably won't be updated, but we add createdAt and updatedAt to all our models as a convention

PR #99, opened 2022-01-06
Point out that Brave browser has issues with browser-based login

sreekaransrinath:


uiuxarghya (approved):

ndneighbor:
Can you elaborate more on the issue that you faced? I don't think this belongs in the docs and would rather have the issue resolved.

FarazPatankar:
Closing as the docs already guide towards using the `--browserless` flag for users who face issues and there hasn't been a response on the PR for a while.

PR #107, opened 2022-02-11
Improved image component and layout

coffee-cup:
- Docs images slightly bleed outside of the container on large screens
- Clicking on an image opens it in a new tab
- Larger area for docs content
- Consistent page nav sidebar

image


ndneighbor (approved):
LGTM

PR #111, opened 2022-02-18
Fix curl URL

aleksrutins:
`railway-develop.app/install.sh` (from #109) gives `ECONNREFUSED`, but `railway.app` seems to work.


JakeCooper:
Oh my. Yes thanks for fixing this!

PR #115, opened 2022-04-01
Document RAILWAY_ENVIRONMENT environment variable

JasonSanford:
I'm running postgres backups from multiple environments in my railway project and was getting some naming collisions and needed a way to differentiate the `dev` backups from `production`, etc. I noticed this `RAILWAY_ENVIRONMENT` was provided but not documented. If this isn't meant to be public or used feel free to 🚮.


coffee-cup (approved):
Thanks!

PR #118, opened 2022-04-03
Remove duplicate build info from ruby page

sqveeze:
The page: https://docs.railway.app/deploy/ruby contains the first part of this page: https://docs.railway.app/deploy/builds

Ruby page starts with the same content like the builds page and can be confusing.


coffee-cup:
Thanks!

PR #124, opened 2022-05-09
Add service flag to 'up' section

zacwebb:
Was looking for a way to specify a service for `railway up` and had to go digging in the cli code to find the `--service` flag added in https://github.com/railwayapp/cli/pull/208/

Cheers 😄


FarazPatankar:
Thanks for adding this, @zacwebb. This is great!

PR #128, opened 2022-05-25
Usecases Doc + Mobile Nav Fix

ndneighbor:
Adds a document on our recommendations on what workloads we recommend for our enterprise users, additionally fixes a pesky bug that wouldn't close the nav after clicking an option on the Mobile Sidebar.

### DIFF
--- a/src/components/Nav.tsx
+++ b/src/components/Nav.tsx
@@ -1,11 +1,12 @@
 import React, { useState } from "react";
+import { useRouter } from "next/router";
 import { ArrowRight, Menu, X } from "react-feather";
 import tw from "twin.macro";
 import { Link } from "./Link";
 import { Logo } from "./Logo";
-import { Search } from "./Search";
-import { MobileSidebar } from "./Sidebar";
+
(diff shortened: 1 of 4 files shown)


src/components/Nav.tsx
```diff
@@ -64,7 +66,41 @@ export const MobileNav: React.FC = () => {
...
+                    >
+                      {page.title}
+                    </Link>
+                  </li>
+                ))}
+              </ul>
+            </React.Fragment>
+          ))}
+        </>
+      </div>
```
coffee-cup, line 103: I'm not sure why the sidebar code is copy and pasted here? I think instead we should be listening for changes on the route and can close the sidebar if it changes.

This change also has the sidebar open when you first visit the page on mobile which isn't ideal.

coffee-cup (changes requested):

src/components/Nav.tsx
ndneighbor: I didn't know I can listen to the page change event. I knew my solution had a smell, but wasn't clever enough to pull that up. I will update.

coffee-cup (approved):

PR #140, opened 2022-07-31
postgredsql.md

CybeeTronic:


treeot:
It's not supposed to have your postgredsql url.

PR #152, opened 2022-09-15
Fix typo in create services documentation

sayopaul:
there were two consecutive "to" 's . I changed it to one.


ndneighbor (approved):

ndneighbor:
Thanks! Good catch :)

PR #167, opened 2022-11-09
Fix double for

Asartea:
Removes a stray for


ndneighbor (approved):
What a catch- typical typo on my part. Thanks 👍

PR #168, opened 2022-11-13
Fix a lot of common grammar mistakes

Milo123459:
- add a note about freenom
- fix grammar mistakes and clarify some things


ndneighbor (approved):
Thanks for reviewing my prose. Will merge in.

PR #177, opened 2022-11-23
fix: link to staticfile builds

lakhansamani:
The link to `staticfile` build has changed and the current URL was redirecting to a 404 page


FarazPatankar (approved):

FarazPatankar:
Thanks for the fix! 😄

PR #184, opened 2022-12-07
Fix typo in README

samuelwstout:
Found a small typo in the README


ndneighbor:
Although it's not a word, we do use the phrase "automagic" for magical experiences on Railway. Appreciate the PR but this will be marked as a "Won't Fix, Not a Bug." :)

PR #194, opened 2022-12-17
Fix typo in deployments.md

gimenete:
What the title says :)


FarazPatankar (approved):
Thanks for the fix!

FarazPatankar (approved):
Thanks for the fix!

PR #196, opened 2022-12-27
Update healthchecks link

lautaropaske:
Old link is no longer in `/diagnose`, page throws 404


FarazPatankar (approved):
Thanks for the fix!

PR #202, opened 2023-01-27
Fresh Coat of Paint to Docs Home Page

ndneighbor:
# What's In This PR

This is a visual update to the Docs to bring the Docs inline with the 2023 Visual Guidelines for Railway. You can find improved use of color, more sophisticated graphics, and new iconography. 

## Changes this:

image

## ... To This

image


src/pages/index.tsx
```diff
@@ -4,55 +4,77 @@ import React from "react";
...
               quality={100}
             />
           </div>
         </div>
 
         <div tw="mt-8 grid grid-cols-1 gap-4 md:grid-cols-2 md:gap-8">
           <OverviewLink href="reference/templates">
-            <OverviewLinkIcon>
-              <Zap size="28" tw="text-pink-800" />
+            <OverviewLinkIcon tw="bg-pink-100 border-pink-100 dark:bg-[#291839]">
```
FarazPatankar, line 74: `dark:bg-[#291839]` are these colors not in the color palette? Looks like we're hardcoding them for all the icons here?

src/styles/colors.ts
```diff
@@ -70,6 +70,11 @@ const lightColors = {
     800: "#0037C7",
     900: "#002994",
   },
+
+  partnerProjectsGradient:
+    "linear-gradient(31.2deg, #0B4347 -25.4%, rgba(9, 36, 45, 0.15) 80.73%), linear-gradient(109.68deg, rgba(60, 0, 61, 0.01) 30.31%, #20041C 85.93%), linear-gradient(149.12deg, #1B0066 13.67%, rgba(28, 0, 108, 0) 86.42%);",
+  submitPartnerProjectGradient:
```
FarazPatankar, line 76: These two should now be removed, right?

FarazPatankar (approved):

src/pages/index.tsx
ndneighbor: Yes, in this case, the color inversion messes up the scale, so I am hard coding them in until we can get to the bottom of that weird color variable not being found in the twin class.

PR #208, opened 2023-02-20
Rewriting docs, making more formal

ItsGoldeneyes:
Hot off of writing a paper for uni, felt like the docs could use some work. First PR of many!

- Changed all references to the starter plan to trial plan. Had a convo with Angelo about this on Discord, figured I'd get ahead of the game.
- Formalized the speech used. Not referring to the user directly
- Rewrote the Heroku Comparison page. Felt like it could use an update, wasn't very convincing. Lots of unnecessary background + praising Heroku. You're equals now! No need to suck up

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -30,7 +30,7 @@ module.exports = withPlugins([
     async redirects() {
       return [
         {
-          source: "/reference/starters",
+          source: "/reference/s",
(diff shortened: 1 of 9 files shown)


ndneighbor (approved):
Much better prose than the one on my end. My writing style is a bit superfluous.

PR #213, opened 2023-03-01
Fix table display

nebulatgs:
The previous PR broke table display on the docs, this PR adds `remark-gfm` to support tables.
Old:
CleanShot 2023-03-01 at 15 06 44@2x

New:
CleanShot 2023-03-01 at 15 07 12@2x


ndneighbor (approved):
👍

PR #215, opened 2023-03-08
use contentlayer instead of next-mdx-remote

nebulatgs:
Uses [Contentlayer](https://www.contentlayer.dev/) to render Markdown instead of `next-mdx-remote`

This allows Hot Reload to work with Markdown changes!


ndneighbor (approved):
That `yarn.lock` regren is really preventing you from code-golfing huh.

LGTM

PR #225, opened 2023-03-27
refactor to a semantic typesense search

neverabsolute:
- Removes fuse.js in favor of typesense, a flexible semantic search engine
- Typesense instance is currently hosted @ [docs-search-index.fly.dev](docs-search-index.fly.dev)

PR #226, opened 2023-03-31
feat(content): Explain how to fix common errors

half0wl:
This PR:
* Creates a new "Troubleshoot" section in the Railway docs
* Add a "Fixing Common Errors" section underneath it
* Add instructions for how to fix "Application Error: This application failed to respond" (i.e. wrong host/port)

## Motivation

In the Discord community, there are a lot of help requests on post-deployment issues. They're commonly user errors unrelated to Railway, and a subset of them are Railway errors unrelated to the platform - for those issues, the problem & solution is always the same.

I believe that a section dedicated to troubleshooting will streamline how Railway provides support. It eases the burden of community helpers & the Railway team by allowing them to link out to highly-specific help sections in the docs without "manually" responding individually to each help request.

More generally, this new section will hold content for troubleshooting common errors users may encounter. Apart from the newly-added page ("Fixing Common Errors"), we can add tech-specific pages into this section that explains common specific gotchas (such as logs not appearing in Python because they're buffered & not flushed to stdout):

```
Troubleshoot
  |> Fixing Common Errors
  |> Python
  |> Node
  |> ...
```

## Rendered Changes

rw-troubleshoot-fixing-common-errors

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -9,6 +9,7 @@ const nextConfig = withContentlayer({
       "railway.app",
       "res.cloudinary.com",
       "devicons.railway.app",
+      "i.imgur.com",
     ],
   },
   async redirects() {
--- a/src/data/sidebar.ts
+++ b/src/data/sidebar.ts
@@ -122,6 +122,23 @@ export const sidebarContent: ISidebarContent = [
     ],
   },
   {
+    title: "Troubleshoot",
+    pages: [
+      makePage("Fixing Common Errors", "troubleshoot", [
+        "error",
+        "errors",
+        "railway error",
+        "railway errors",
+        "application error",
+        "failed to respond",
+        "application failed to respond",
+        "503",
+        "service unavailable",
+        "503 service unavailable",
+      ]),
+    ],
+  },
+  {
     title: "Reference",
     pages: [
       makePage("Pricing", "reference", ["pricing"]),
--- a/src/docs/troubleshoot/fixing-common-errors.md
+++ b/src/docs/troubleshoot/fixing-common-errors.md
@@ -1,0 +1,102 @@
+---
+title: Fixing Common Errors
+---
+
+{/**
+This page holds descriptions and solutions to common errors users may face
+when deploying to Railway.
+
+When adding a new error/section, please keep in mind that content on this
+page should be as detailed as possible, and link out to relevant docs when
+necessary. We want to be consistent and complete. When in doubt, try to
+stick to this formula:
+
+    - Error as section title (`## Some Error`)
+        - A description of the error. Explain to users what error this is (add
+          a screenshot if applicable), why this happens, and what causes it.
+        - How to solve the error (`## Solution`). Solutions may have sections
+          underneath it for language/framework/stack-specific solutions, e.g.:
+            - Python
+            - Go
+            - ...
+*/}
+
+When deploying to Railway, you may encounter some errors that prevent your
+application from working. These are descriptions and solutions to errors that
+users commonly encounter.
+
+## Application Error: This application failed to respond
+
+After deploying your application, you encounter this screen when accessing
+your application's domain:
+
+<Image src="https://i.imgur.com/4WDZHHE.png"
+alt="Screenshot of application failed to respond error"
+width={729} height={675}
+quality={80} />
+
+This error occurs when Railway is unable to connect to your application.
+
+Railway needs to know how to communicate with your application. When you
+deploy and expose a web application on Railway, we expect your web server
+to be available at host `0.0.0.0` and a port that we provide in the form
+of a `PORT` variable. The `PORT` variable is automatically injected by
+Railway into your application's environment.
+
+Thus, your web server must listen on host `0.0.0.0` and the port that
+Railway provides in the `PORT` environment variable.
+
+### Solution
+
+To fix this, start your application's server using:
+
+* Host = `0.0.0.0`,
+* Port = Value of the `PORT` environment variable provided by Railway.
+
+<Banner variant="info">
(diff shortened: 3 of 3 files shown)


next.config.js
```diff
@@ -9,6 +9,7 @@ const nextConfig = withContentlayer({
       "railway.app",
       "res.cloudinary.com",
       "devicons.railway.app",
+      "i.imgur.com",
```
half0wl, line 12: Needs to be removed once I get a Cloudinary URL to the image.

src/docs/troubleshoot/fixing-common-errors.md
```diff
@@ -0,0 +1,102 @@
...
+When deploying to Railway, you may encounter some errors that prevent your
+application from working. These are descriptions and solutions to errors that
+users commonly encounter.
+
+## Application Error: This application failed to respond
+
+After deploying your application, you encounter this screen when accessing
+your application's domain:
+
+<Image src="https://i.imgur.com/4WDZHHE.png"
```
half0wl, line 33: (should not be on Imgur)

Can someone help me upload this to Cloudinary and paste the link here? I'll update this and remove `i.imgur.com` from allowed the allowed images domain once that's done. Thanks!

src/docs/troubleshoot/fixing-common-errors.md
```diff
@@ -0,0 +1,102 @@
...
+    ```
+
+* **Flask**
+    ```sh
+    # Procfile
+    web: gunicorn --bind=0.0.0.0:$PORT $YOUR_APP:app
+    ```
+
+For more information about Procfiles, check out
+[Builds -> Procfiles](/deploy/builds#procfiles).
```
half0wl, line 102: I'm a bit unsure about this section - `gunicorn` already works by default, so this may be entirely unnecessary and cause further confusion.

Thoughts?

src/data/sidebar.ts
```diff
@@ -121,6 +121,23 @@ export const sidebarContent: ISidebarContent = [
...
+        "errors",
+        "railway error",
+        "railway errors",
+        "application error",
+        "failed to respond",
+        "application failed to respond",
+        "503",
+        "service unavailable",
+        "503 service unavailable",
+      ]),
```
half0wl, line 138: Went a bit too verbose on this. I want to make sure this is reachable via search.

This can go away after #225 lands

src/docs/troubleshoot/fixing-common-errors.md
```diff
@@ -0,0 +1,102 @@
...
+
+* Host = `0.0.0.0`,
+* Port = Value of the `PORT` environment variable provided by Railway.
+
+<Banner variant="info">
+Alternatively, you can set a custom `PORT` service variable in your
+Railway environment to tell Railway that your application is available
+at the port you specified. For more information, check out
+[Defining Variables](/develop/variables#defining-variables).
+**This approach is not recommended**.
```
half0wl, line 61: Maybe explain why it's not recommended?

src/docs/troubleshoot/fixing-common-errors.md
half0wl, line 33: Unrelated suggestion: Link out to this page on the "Application failed to respond" iframe :-)

src/docs/troubleshoot/fixing-common-errors.md
aleksrutins: ... because it makes your code more tied to your configuration and therefore makes your project less transferable, maybe? Just a guess.

src/docs/troubleshoot/fixing-common-errors.md
```diff
@@ -0,0 +1,102 @@
...
+  // ...
+});
+```
+
+#### Python/Gunicorn
+
+`gunicorn` listens on `0.0.0.0` and the `PORT` environment variable by default.
+There is no additional configuration necessary.
+
+If you are running into this error on `gunicorn`, try creating a
```
aleksrutins, line 83: I'd suggest using `nixpacks.toml` instead of a Procfiles, as Railway's Procfiles are slightly different from normal Heroku Procfiles.

For Django:
```toml
[start]
cmd = "python manage.py migrate && python manage.py collectstatic --noinput && gunicorn --bind=0.0.0.0:$PORT $YOUR_APP.wsgi"
```
For Flask:
```toml
[start]
cmd = "gunicorn --bind=0.0.0.0:$PORT $YOUR_APP:app"

aleksrutins (approved):
Generally, LGTM; the one thing that I'd suggest is to use `nixpacks.toml` over a Procfile.

src/docs/troubleshoot/fixing-common-errors.md
half0wl: Any thoughts on removing this section altogether? I think it could potentially confuse users [0] since gunicorn already does this by default

[0] https://discord.com/channels/713503345364697088/1088240586768797756/1091524760434901102

src/docs/troubleshoot/fixing-common-errors.md
half0wl: Yep! Overriding the default behaviour for no good reason might also introduce more unexpected issues, generally speaking. I think it's a sane default, and it's there for a reason.

Just wondering if that should belong in here or link out to a separate section dedicated to informational content about how this works.

half0wl (commented):
https://docs.railway.app/deploy/deployments#bad-gateway should also be updated in light of this change. 502 Bad Gateway is also not the error for this (the error is a 503 unavailable), so I think that section is incorrect/outdated.

half0wl:
[resolves log-282]

PR #240, opened 2023-04-24
feat(search)!: Migrate to new Search UI powered by Meilisearch

half0wl:
This replaces the existing tag-based search with site-wide full-text search, powered by Meilisearch Cloud.

* Index is built using a manually-triggered GH action introduced in #235 ([actions/build-search-index](https://github.com/railwayapp/docs/actions/workflows/build-search-index.yml) -> Run Workflow).
* Results are organized by the hierarchy root: each section in the sidebar maps to a section in results (e.g. "Deploy") with individual results as its children.
* Re-usability is a key design consideration. We plan to add this to [nixpacks docs](https://nixpacks.com/docs). The components introduced here contains as little logic as possible, with the core functionality exposed by a `useDebouncedSearch` hook. 

### Live demo

[railway-docs-docs-pr-240.up.railway.app](https://railway-docs-docs-pr-240.up.railway.app/)

### Related PRs
* #242 
* #241

## Screens
ss 2023-04-24 at 8 13 17 PM

ss 2023-04-24 at 8 13 32 PM

## Future Improvements
* Migrate from Meilisearch Cloud -> Meilisearch on Railway
  * Needs persistent vol. While the index gets nuked on each build, data going away on re-deploys or restarts is a recipe for "nothing shows up in search" class of issues until the build gets triggered again
* Auto-trigger index build on each deploy
  * Not worth doing a custom build for chaining actions right now
* Add back hotkey navigation
  * Removed this in favour of moving faster. Definitely a UX regression that should be taken care of.

## Rollout
- [ ] Add new env vars to `production`:
  - [ ] `NEXT_PUBLIC_MEILISEARCH_HOST`
  - [ ] `NEXT_PUBLIC_MEILISEARCH_READ_API_KEY`
  - [ ] `NEXT_PUBLIC_MEILISEARCH_INDEX_NAME`
- Ideal landing order:
  1. #240
  2. #242
  3. #241


src/components/Search/Results.tsx
```diff
@@ -0,0 +1,102 @@
...
+    return acc;
+  }, {} as Result);
+
+  // @FIXME: Indexer is grabbing #__next from anchor hrefs. This should be
+  // fixed upstream, but no harm in just hacking it in place for now.
+  const cleanSlug = (slug: string) => slug.replace("#__next", "");
+
+  return (
+    <div css={tw`p-2 m-2`}>
+      {Array.from(Object.entries(results)).map(([chapter, hits]) => {
```
coffee-cup, line 57: Not sure if the `Array.from` is necessary as `Object.entries` should already return an array that can be mapped over

src/components/Search/Results.tsx
```diff
@@ -0,0 +1,102 @@
...
+          slug: hit.url,
+          text: (hit._formatted && hit._formatted.content) ?? "",
+        })),
+    ];
+    return acc;
+  }, {} as Result);
+
+  // @FIXME: Indexer is grabbing #__next from anchor hrefs. This should be
+  // fixed upstream, but no harm in just hacking it in place for now.
+  const cleanSlug = (slug: string) => slug.replace("#__next", "");
```
coffee-cup, line 53: I am noticing that clicking on links in the search results brings you to the production site in a new tab (docs.railway.app). I think we should stripe out the full domain and only link to the the subpath. This would also allow us to take better advantage of NextJS client side routing.

src/hooks/useDebouncedSearch.ts
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      const response = await index.search<T>(query, params);
+      setResponse(response);
+    } catch (e) {
+      console.error(`Search for query "${query}" failed (${e})`);
+    }
+  }, [query]);
+
+  // Perform search
+  useEffect(() => {
```
coffee-cup, line 51: Suggestion: Use react-query here instead of `useEffect` as it will handle loading/error states better, as well as has a few QoL features for fetching data from APIs.

coffee-cup (commented):
Overall looks good! I think there should be a follow up PR that fixes the keyboard navigation since that was pretty confusing when I first tried it out.

There also isn't a clear separation between the result items, but that is maybe a Jitachi question?

src/components/Nav.tsx
```diff
@@ -4,6 +4,7 @@ import { Home, Menu, X } from "react-feather";
 import tw from "twin.macro";
 import { Link } from "./Link";
 import { Logo } from "./Logo";
+import { OpenSearchModalButton } from "@/components/Search";
```
ndneighbor, line 7: Super duper tiny nit: but we may wanna standardize the way we do imports on this repo moving forward. Maybe best as an incoming todo.

src/components/Search/Modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+const Modal: React.FC<Props> = ({ closeModal }) => {
+  const searchParams = {
+    limit: 10,
+    attributesToHighlight: ["*"],
+    highlightPreTag: "<span>",
+    highlightPostTag: "</span>",
+  };
+  const { clearResponse, query, setQuery, response } =
+    useDebouncedSearch<Search.Document>(
+      process.env.NEXT_PUBLIC_MEILISEARCH_HOST ?? "",
```
ndneighbor, line 22: Nit: Hmmm, we can use something like `zod` to make it so that the access to the env vars is confirmed to be in the env before referencing it in the front-end.

ndneighbor (approved):
Nothing else to add except from JR's comments. I have two nits, but mostly they are stylistic concerns that can be waived away if you don't feel that they are important.

src/components/Nav.tsx
half0wl: Yeah. I've defaulted to `@/...` in this PR, but didn't want to touch anything irrelavent to reduce churn.

src/components/Search/Results.tsx
```diff
@@ -0,0 +1,100 @@
...
+            hit.hierarchy_lvl2,
+            hit.hierarchy_lvl3,
+            hit.hierarchy_lvl4,
+          ].filter(h => h !== null),
+          slug: hit.url,
+          text: (hit._formatted && hit._formatted.content) ?? "",
+        })),
+    ];
+    return acc;
+  }, {} as Result);
```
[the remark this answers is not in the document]
half0wl: Good idea!

Added in 82a28783572c0c1e75e984879e287ef9dbba801e, though that `Transformer<any, any>` line is quite icky. Dunno if it's worth doing this "the right way," I guess it's fine as long as the response and result type makes it back to the caller.

src/components/Search/Results.tsx
half0wl: Fixed in c016b1501b6f4b37def36fcff62921540ce93387

src/components/Search/Results.tsx
half0wl: 82a28783572c0c1e75e984879e287ef9dbba801e

src/components/Search/Modal.tsx
half0wl: Good idea. I'd like to do this in a separate pull, and switch this over in there.

src/components/Nav.tsx
half0wl: #244

src/components/Search/Modal.tsx
half0wl: #245

coffee-cup (approved):

PR #241, opened 2023-04-24
feat(analytics): Track search events

half0wl:
This tracks a new "Search Performed" goal in Fathom analytics. **No search query is sent! (Fathom doesn't even support it)**

While the cloud offering of Meilisearch has basic "analytics" that tells us how many searches are performed, it'd be nice to have this tracked over time.

Depends on #240.


ndneighbor (approved):
Stamping with approval but I consider this a premature split from the main PR. This would be nice to have to the main branch when you merge in the above.

half0wl:
> Stamping with approval but I consider this a premature split from the main PR. This would be nice to have to the main branch when you merge in the above.

They're split for easier reviews, testability, and easy reverts (generally speaking). They will be going out at the same time :-)

PR #251, opened 2023-04-26
feat(content): Refresh Develop -> CLI content

half0wl:
This PR does a slight refresh of https://docs.railway.app/develop/cli to re-organize information in a reader-friendlier way, and also fixes the following issues:
- `...Open an interactive shell to a database within a project...` False, we don't support this (it's a shell with environment variables, not a shell to a database)
- `...Homebrew (Linux, macOS)...` Homebrew is not available on Linux
- CLIv3
  - `railway shell` is not removed
  - `railway down` is not removed

### DIFF
--- a/src/docs/develop/cli.md
+++ b/src/docs/develop/cli.md
@@ -7,84 +7,122 @@ alt="GIF of the CLI in Action"
 layout="intrinsic"
 width={800} height={468} quality={100} />
 
-The Railway command line interface (CLI) connects your code to your Railway project from the command line.
+The Railway Command Line Interface (CLI) lets you interact with your
+Railway project from the command line, allowing you to:
 
-The Railway CLI allows you to
+- Create new Railway projects from the Terminal
+- Link a local project directory to an existing Railway project
+- Run services locally using environment variables from your Railway project
+- And more!
 
-- Create new Railway projects from the terminal
-- Link to an existing Railway project
-- Pull down environment variables to run a service locally
-- Open an interactive shell to a database within a project
+## Installation
 
-## Install
+The Railway CLI can be installed via:
 
-Install with [Brew](https://brew.sh), [NPM](https://www.npmjs.com/package/@railway/cli), Shell Script, or [Scoop](https://scoop.sh).
+- [Homebrew](https://brew.sh) for macOS
+- [npm](https://www.npmjs.com/package/@railway/cli) for macOS, Linux, and Windows
+- [Scoop](https://scoop.sh) for Windows as a native `.exe` binary
+- Direct Shell Script
+- Downloading pre-built binaries
+- Building from source
 
-### NPM (Linux, macOS, Windows)
+### Homebrew (macOS)
 
-```bash
-npm i -g @railway/cli
-```
-
-You need to have => 16.x version of [Node.js](https://nodejs.org/en/) installed.
-
-### Shell Script (Linux, macOS, Windows: Git Bash)
-
-```bash
-curl -fsSL cli.new | sh
-```
-
-### Homebrew (Linux, macOS)
+In a Terminal, enter the following command:
 
 ```bash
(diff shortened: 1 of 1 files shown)


src/docs/develop/cli.md
```diff
@@ -7,84 +7,122 @@ alt="GIF of the CLI in Action"
...
-If there are any issues with browser-based login you can use the `--browserless` flag to authenticate.
+This command opens a new tab in your default browser to the [https://railway.app](https://railway.app)
+authentication page. Follow the instructions to complete the authentication process.
+
+### Manual Login
 
-This will print a URL and a pairing code to the terminal which you can use to authenticate (useful for SSH sessions).
+You can also authenticate manually using a Pairing Code. This can be useful if
+you're authenticating the CLI inside an environment without a browser (e.g. SSH
+sessions).
```
ndneighbor, line 96: No review comment but just pointing out what I like about this PR. This is clear, explains benefit, and an improvement on the curt nature of the copy.

ndneighbor (changes requested):
Address the Homebrew point but I believe the rest is great!

src/docs/develop/cli.md
```diff
@@ -180,31 +219,31 @@ Our CLI is open source. Contribute to the development of the Railway CLI by open
...
 
-- Undocumented commands such as `railway protect` and `railway shell` will be removed. We are re-thinking how we want to handle these features and will be adding them back in the future
+- `railway build` has been removed
+- `railway protect` has been removed
 - `railway variables add` will no longer give you the option to add a secret. We are re-thinking how we want to handle the secret flow within the CLI now since the addition of [Shared Variables](/develop/variables#shared-variables)
-- `railway down` will be sunset as we are looking into better ways to handle deployments and scaling
-- `railway version` is being removed in favor of using `-V` or `--version` flags.
-- `railway build` is being removed in favor of using `railway variables` to output a local .env file (also the name was confusing)
+- `railway version` has been removed, in favor of using `-V` or `--version` flags
 - `railway init` will no longer allow you to deploy a new Project with a Template. We have since re-worked the flow and realized that the Template deploy flow is best served in the dashboard
```
ndneighbor, line 249: We may wanna eventually stage this for removal after the CLI regressions project is completed.

src/docs/develop/cli.md
```diff
@@ -7,84 +7,122 @@ alt="GIF of the CLI in Action"
...
-Install with [Brew](https://brew.sh), [NPM](https://www.npmjs.com/package/@railway/cli), Shell Script, or [Scoop](https://scoop.sh).
+- [Homebrew](https://brew.sh) for macOS
+- [npm](https://www.npmjs.com/package/@railway/cli) for macOS, Linux, and Windows
+- [Scoop](https://scoop.sh) for Windows as a native `.exe` binary
+- Direct Shell Script
+- Downloading pre-built binaries
+- Building from source
 
-### NPM (Linux, macOS, Windows)
+### Homebrew (macOS)
```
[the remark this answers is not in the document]
half0wl: > I don't think this is entirely inaccurate. With that said, maybe we should link to the `brew` docs for distros that we don't support.
> 
> https://docs.brew.sh/Homebrew-on-Linux

...interesting! Is brew a popular package manager on Linux? While it's technically true and accurate, I've never seen it recommended/specified as a installation method anywhere. 

> ...Plus maybe caveat by saying `WSL` instead of Windows.

4a073dbcade0fd5e5a82cbfebe68c21fff783095

src/docs/develop/cli.md
half0wl: Yep, I wanted to remove it entirely, but v3 isn't exactly "done" until regressions are taken care of. We can slate this for 23Q3.

ndneighbor (approved):

Issue #249, opened 2023-04-25
Search UI design refresh


half0wl:
The search modal currently looks bare. Going to collab w/ @jitachi on this in the coming weeks.

Related to #248, might want a dedicated page we can mount as a modal or something along those lines.

PR #255, opened 2023-04-27
feat(search): Add hotkey nav, tweak UI slightly for better clarity between results

### LINKED ISSUE #247: Add hotkey nav to Search UI
UX regression from #240.

half0wl:
This adds back hotkey-based nav in the search UI:
* Arrow keys (`^`, `v`) to navigate through results
* `Enter` to browse to selected result
* Hover to select result

UI tweaks:
* Make results stand out more by adding borders
* Clip text overflow for long sections, e.g.
    ss 2023-04-28 at 3 10 19 AM

## Demo

https://user-images.githubusercontent.com/24421625/234967059-addad8db-8ccd-4e27-9f9f-abbdfa12a38d.mov

Resolves #247
Resolves LOG-351

### DIFF
--- a/src/components/Search/Results.tsx
+++ b/src/components/Search/Results.tsx
@@ -68,7 +68,7 @@ const Results: React.FC<Props> = ({ closeModal, results }) => {
     }
     closeModal();
     router.push(selectedResult.slug);
-  }, [selectedResult]);
+  }, [closeModal, router, selectedResult]);
 
   useEffect(() => {
     const unsubscribe = tinykeys(window, {
@@ -77,7 +77,7 @@ const Results: React.FC<Props> = ({ closeModal, results }) => {
       Enter: () => onEnter(),
     });
     return () => unsubscribe();
-  }, [onArrowKeyDown]);
+  }, [onArrowKeyDown, onArrowKeyUp, onEnter]);
 
   return (
     <div css={tw`p-2 m-2`}>
--- a/src/components/Search/Results.tsx
+++ b/src/components/Search/Results.tsx
@@ -1,25 +1,84 @@
 import { Link } from "@/components/Link";
 import { Search } from "@/types";
 import { Markup } from "interweave";
-import React from "react";
+import { useRouter } from "next/router";
+import React, { useCallback, useEffect, useState } from "react";
 import tw from "twin.macro";
+import tinykeys from "tinykeys";
 
 interface Props {
-  closeModal: () => void
+  closeModal: () => void;
   results: Search.Result;
 }
 
-const Results: React.FC<Props> = ({ closeModal, results }) => {
-  const withoutBaseUri = (slug: string) => {
-    const url = new URL(slug);
-    const { hash, pathname } = url;
-    return `${pathname}${hash}`;
-  };
+type SelectedResult = { idx: number; slug: string };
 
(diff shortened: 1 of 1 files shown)


src/components/Search/Results.tsx
```diff
@@ -1,25 +1,83 @@
...
 import { Markup } from "interweave";
-import React from "react";
+import { useRouter } from "next/router";
+import React, { useCallback, useEffect, useState } from "react";
+import tinykeys from "tinykeys";
 import tw from "twin.macro";
 
 interface Props {
-  closeModal: () => void
+  closeModal: () => void;
```
half0wl, line 10: (prettier churn, ignore)

src/components/Search/Results.tsx
```diff
@@ -1,25 +1,83 @@
...
+
+const withoutBaseUri = (slug: string) => {
+  const url = new URL(slug);
+  const { hash, pathname } = url;
+  return `${pathname}${hash}`;
+};
+
+// @FIXME: Indexer is grabbing #__next from anchor hrefs. This should be
+// fixed upstream, but no harm in just hacking it in place for now.
+const cleanSlug = (slug: string) => withoutBaseUri(slug.replace("#__next", ""));
```
half0wl, line 24: (slight refactor; moved out of component, ignore)

src/components/Search/Results.tsx
```diff
@@ -1,25 +1,83 @@
...
+    });
+  }, [resultsFlat, selectedResult, setSelectedResult]);
+
+  const onArrowKeyUp = useCallback(() => {
+    setSelectedResult(prev => {
+      if (prev === null || prev.idx === 0) {
+        // Start of results. Going up from here is the search input, so it's
+        // a no-op.
+        return null;
+      }
```
half0wl, line 58: This would ideally also tweak the input focus:
* on key down, remove focus on input
* on key up to here (idx=0/head of result), re-focus input

Requires some `ref` trickery to wire up. I don't want to get into that right now. Also may get somewhat confusing since it requires different stateful props to be passed around (between Modal, QueryInput, Results I think), so it's best to do that in a separate pull and keep this lean.

src/components/Search/Results.tsx
```diff
@@ -36,27 +94,54 @@ const Results: React.FC<Props> = ({ closeModal, results }) => {
...
+                  >
                     <Link
-                      href={cleanSlug(h.slug)}
-                      onClick={(e) => closeModal()}
-                      css={[tw`flex flex-col font-medium hover:text-pink-700`]}
+                      href={slug}
+                      onClick={closeModal}
+                      css={[
+                        tw`flex flex-col font-medium p-3 rounded rounded-lg border`,
+                        isSelected && tw`bg-pink-200`,
```
half0wl, line 119: Staying in-line with the current sidebar design. Would like to re-visit in #249

src/components/Search/Results.tsx
ndneighbor: I agree here, best to have the listener like this until we can do something more ideal. Good call.

ndneighbor (approved):
Really solid work, really appreciate the fast turnaround on this.

coffee-cup (commented):
I see a application error when pressing cmd+k in the preview

image

Issue #250, opened 2023-04-25
Anchor links sometime do not take you directly to the section it belongs to


half0wl:
I can't reproduce this reliably, but go to https://docs.railway.app/reference/cli-api and click around anchor links and you'll land on a different section than intended after a few clicks.

This seems related to https://github.com/vercel/next.js/issues/44295

artimath:
@half0wl this is likely unrelated to this specific issue, and the marketing site is likely a different repo... 

but worth noting here that the "main career page" links in job pages link to the same job page. reported with screen record in [discord](https://discord.com/channels/713503345364697088/872280333167517736/1157420137377452084). happens on both safari ios and chrome mac for me.

<img width="1728" alt="2023-09-29_11-39-53" src="https://github.com/railwayapp/docs/assets/19831746/d6bc9346-9c49-42d7-9680-2df861c35340">

artimath:
Also, can confirm that the original issue here happens still on my machine. I might take a look and report back.

Teejay128:
Hello all, I think this issue has been resolved, or at least I haven't been able to reproduce the issue described above by clicking on the links highlighted in the image below.
![image](https://github.com/user-attachments/assets/83633c75-c3ee-4671-bdf1-94d111087456)

brody192:
This was fixed in #558

Issue #246, opened 2023-04-25
Add anchor tags to section headers


half0wl:
https://cdn.discordapp.com/attachments/1100364112862646332/1100364494372352080/CleanShot_2023-04-25_at_03.15.192x.png

This allows us to link out to any section without scrolling up to the ToC sidebar on the right.

samridht23:
[railwayfix.webm](https://user-images.githubusercontent.com/75132459/235430045-77a62440-7a09-41d4-9dcf-9737982df4e8.webm)

i made a pr that will add the (#) fragment-identifier to the titles but i was not able to overwrite the span that which was already present there.
 
![railwayfix](https://user-images.githubusercontent.com/75132459/235430838-4cfb6eff-6a96-4b49-8061-9550840d0140.png)

half0wl:
Closed in #261

PR #261, opened 2023-05-01
feat(ui): Add anchor link to headings

samridht23:
added fragment identifier to titles so we can directly link the part of the docs and do not have to scroll the to find a particular heading

### DIFF
--- a/src/pages/[...slug].tsx
+++ b/src/pages/[...slug].tsx
@@ -48,67 +48,64 @@ const StyledHeadingH2 = styled.h2`
 `;
 
 const Image = (props: ImageProps) => (
-    <a
-        tw="block xl:-mx-8"
-        href={props.src as string}
-        target="_blank"
-        rel="noopener"
-    >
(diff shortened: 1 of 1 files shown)


half0wl (changes requested):
Thank you for the PR!

This throws an error for me locally:
```
Error: Hydration failed because the initial UI does not match what was rendered on the server.
See more info here: https://nextjs.org/docs/messages/react-hydration-error
```

Do you happen to know why?

It also looks like anchors are getting nested:

ss 2023-05-04 at 12 53 19 AM

samridht23:
I have addressed the hydration issue and removed the nested <a> tag in this pull request. That was caused by improper handling of server-side rendering and hydration of the component. Can you check if the issue still persist.

half0wl (approved):
Works, thank you!

half0wl:
Resolves #250

half0wl:
Resolves #246

PR #276, opened 2023-06-06
feat: add an example on how to use the railway_service env

ThallesP:
It's kinda confusing if the `{}` is required or not, added an example to make it clear


half0wl (approved):
@ThallesP thank you!

Issue #266, opened 2023-05-09
Clear search results when query input is empty


half0wl:
1. Enter some text
2. Delete it
3. Query results are still present, it should not be

Suggested by @brody192

jobayer12:
@half0wl 
Can anyone give some instructions on how to setting up meilisearch? Basically, I'm new here. I can't set up this project properly in the local environment. It would be helpful if anyone gives some instructions for setting up this project in the local environment.

tarunsamanta2k20:
@half0wl Unhandled Runtime Error
MeiliSearchError: The provided host is not valid.

TruePadawan:
I've addressed this in this [pull request](https://github.com/railwayapp/docs/pull/277)

half0wl:
Fixed in #277 by @TruePadawan

PR #277, opened 2023-06-06
Clear search results if query is empty

TruePadawan:
Like this [issue](https://github.com/railwayapp/docs/issues/266) describes, search results were not being removed when the input text is empty.

I fixed that.


half0wl (approved):
Thanks!

PR #280, opened 2023-06-11
Fixed (possible) spelling mistake in plans.md

WilliamPlays0402:
fiat => flat

### DIFF
--- a/src/docs/reference/plans.md
+++ b/src/docs/reference/plans.md
@@ -117,7 +117,7 @@ Credits can be bought with:
 Credits cannot be bought with:
 
 - PayPal - neither PayPal balance nor linked cards on PayPal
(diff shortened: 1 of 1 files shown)


half0wl:
@WilliamPlays0402 thanks for the PR! This is actually not a typo -- fiat, in this context, refers to any official money (of a country) that can be used for barter (such as USD, EUR). It's called fiat because it's controlled by a central authority [0], while crypto is decentralized and not controlled by a single entity.

In here, the sentence is basically saying "real money" like USD vs "crypto" like BTC -- it's a popular term in crypto circles commonly used to disambiguate between the two, so that makes it easier for readers to digest.

More on this if you're curious: https://calbizjournal.com/understanding-crypto-vs-fiat-currency/

[0] https://www.merriam-webster.com/dictionary/fiat

PR #283, opened 2023-06-21
Nonstopflights query doc update

nonstopflights:
Not 100% sure this is the intended query but the example on the site does not work when copy and pasted directly. If not correct, hopefully this serves to bring the issue to light. Thanks!


coffee-cup (approved):
Thanks for the fix! :D

PR #316, opened 2023-08-04
Clarify when cron jobs can be skipped

gimenete:


src/docs/reference/cron-jobs.md
```diff
@@ -86,4 +86,4 @@ Let's break down each field:
 
 Scheduled services should exit as soon as they are done with the task they are responsible to perform. Thus, the process should close any connections, such as database connections, to exit properly.
 
-At this moment Railway won't terminate your process in any scenario and it's not going to skip any execution. Thus, it is possible that if your service is still running when a new invocation is expected, there will be two instances of the service running at the same time. However this may change in the future depending on the feedback we receive.
+At this moment Railway won't terminate your process in any scenario. However cron triggers may be skipped if building and deploying a service takes so long that overlaps with the next scheduled execution.
```
half0wl, line 89:
```suggestion
At this moment, Railway won't terminate your process in any scenario. However, cron jobs may be skipped if the service is in the build/deploy stage when the next scheduled execution happens.
```
^ thoughts on this tiny rephrase?

gimenete: Perfect. Thanks!

PR #323, opened 2023-08-23
Update railway provided variables

coffee-cup:
- Update the Railway provided variables to be up to date
- Split Railwy provided Git variables into separate section
- Add docs on referencing other service Railway provided variables


half0wl (approved):
🤩

PR #337, opened 2023-09-12
Search: Add loading indicator

half0wl:
This adds a loading indicator to the search modal:
cs-2023-09-13-02 47


src/components/LoadingIndicator.tsx
```diff
@@ -0,0 +1,27 @@
+import tw from "twin.macro";
```
FarazPatankar, line 1: Just realized docs still uses twin. 🙈

src/hooks/useDebouncedSearch.ts
```diff
@@ -104,13 +106,16 @@ export const useDebouncedSearch = <
...
     }
     try {
+      setIsSearching(true);
       const response = await index.search<Response>(query, params);
       trackGoal(FATHOM_SEARCH_PERFORMED_EVT_ID, 0);
       setResults(transformResponse(response));
+      setIsSearching(false);
     } catch (e) {
       console.error(`Search for query "${query}" failed (${e})`);
+      setIsSearching(false);
```
FarazPatankar, line 116: Can move this and the one on 113 outside the `catch` to only call it once!

FarazPatankar (approved):

src/components/LoadingIndicator.tsx
half0wl: ye this kinda threw me off..

src/hooks/useDebouncedSearch.ts
half0wl: 32a4b010b7a3d5b76bec8ff5af8e63db634fead5

FarazPatankar: ty ty, pre-stamped btw.

PR #350, opened 2023-09-28
update database docs

melissa-hale:
added a new doc under databases to show that we can support any database

update the existing docs to reflect changes to DB services


melissa-hale:
polished and links checked, looks good

PR #356, opened 2023-10-06
scale to zero doc

melissa-hale:
add doc for scale to zero


src/data/sidebar.ts
```diff
@@ -84,6 +84,7 @@ export const sidebarContent: ISidebarContent = [
       makePage("Volumes", "reference"),
       makePage("Cron Jobs", "reference"),
       makePage("Pricing", "reference"),
+      makePage("Scale to Zero", "reference"),
```
half0wl, line 87: Thoughts on renaming this from "Scale to Zero" to "App Sleeping" entirely?

We use the term "App Sleeping" when exposing this feature, and "Scale to Zero" is more of a "marketing" artifact than a feature - I think we should stay consistent in naming features, and only use a single name everywhere except marketing content that tells a story about what the feature enables you to do

I think it might also be confusing reading this from top-down going from "scale to zero" to "app sleeping" because it leaves too much effort to the reader to figure out the correlation

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Scale to Zero
+---
+
+Scale to Zero means reducing unnecessary costs by ensuring you only use what you need, when you need it.
```
half0wl, line 5: I feel this sentence is way too broad and open to interpretation.

Suggestion - start with:
> App Sleeping allows you to increase the efficiency of resource utilization on Railway. It may reduce the resource usage cost of a [service](/develop/services) by ...

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+
+## What is App Sleeping?
+
+App sleeping enables automated detection of an inactive [service](/develop/services) based on outbound traffic.
+
+Enabling App Sleep on a service tells Railway to stop the service when it is inactive, effectively reducing the overall cost to run it.
```
half0wl, line 11: Suggestion: Replace the sentence above ("App sleeping enables automated ...") with the second one ("Enabling App Sleep on a service ..."), so you get a pyramid style of "what it does" -> "how it does it" (i.e. it tells railway to stop the service -> it does this by detecting traffic)

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+
+<Image src="https://res.cloudinary.com/railway/image/upload/v1696017787/events_gjkaob.png"
+alt="App Sleeping in the Activity Feed"
+layout="intrinsic"
+width={700} height={460} quality={100} />
+
+### Inactive Service Detection
+
+Inactivity is based on outbound packets. If no packets are sent from the service for over 10 minutes or longer, the service is considered inactive.
```
half0wl, line 20: Thoughts on clarifying what types of traffic this applies to?

It's both inbound/outbound and it may be confusing to only reference a single one

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+### Reactivation
+
+When the service receives traffic from the _public network*_, it is sent a signal to wake.
```
half0wl, line 25: Also happens when a service sends traffic to WAN

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+alt="App Sleeping in the Activity Feed"
+layout="intrinsic"
+width={700} height={460} quality={100} />
+
+### Inactive Service Detection
+
+Inactivity is based on outbound packets. If no packets are sent from the service for over 10 minutes or longer, the service is considered inactive.
+
+
+### Reactivation
```
half0wl, line 23: Suggestion: "Waking a Service Up"

If we call this "app sleeping" we should document the states it can be in, in a consistent manner (asleep or awake)

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+
+When the service receives traffic from the _public network*_, it is sent a signal to wake.  
+
+Only when the service is ready, is traffic routed to it, ensuring no requests are dropped during the startup process.
+
+_*Currently, requests made to the service over the private network do not cause it to awake from sleep._
```
half0wl, line 29: Suggestion:
```suggestion
We wake a service when it sends or receives traffic to/from the internet.

 The first request made to or sent from a service wakes a sleeping service up. This may take a small amount of time for the service to spin up again on the first request (commonly known as "cold boot time").
```

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+
+## How to enable App Sleeping
+
+To enable App Sleeping simply toggle the feature on within the service configuration -
```
half0wl, line 33:
```suggestion
To enable App Sleeping, toggle the feature on within the service configuration pane in your project:
```

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+<Image src="https://res.cloudinary.com/railway/image/upload/v1696548703/docs/scale-to-zero/appSleep_ksaewp.png"
+alt="Enable App Sleep"
+layout="intrinsic"
+width={700} height={460} quality={100} />
+
+1. Navigate to your service's settings > Deploy > App Sleeping.
+2. Toggle "Enable App Sleeping".
+3. This setting will be applied across all [replicas](https://docs.railway.app/develop/services#horizontal-scaling-with-replicas) automatically.
+4. When your service is put to sleep, an event will be visible in the dashboard.
+5. To _disable_ App Sleeping, simply toggle the setting again.
```
half0wl, line 44:
```suggestion
1. Navigate to your service's Settings > Deploy > App Sleeping
2. Toggle "Enable App Sleeping"
3. To _disable_ App Sleeping, toggle the setting again
```

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Things to Note
```
half0wl, line 46:
```suggestion
## Caveats
```

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+- **Latency** - There may be a small delay when requests arrive at a stopped application.
+- **Private Networking** - Services that receive traffic exclusively over the private network should *not* be enabled with App Sleeping since only public network traffic will trigger the service to wake.
+- **Conditions for Sleeping** - For a service to be considered inactive, it must not send outbound traffic for at least 10 minutes or more.  Outbound traffic can include telemetry, NTP, etc.
```
half0wl, line 49:
```suggestion
- There will be a small delay in the response time of the first request sent to a slept service (commonly known as "cold boot times")
- [Private Networking](/reference/private-networking) does not currently work with App Sleeping
- Outbound traffic is excluded from considering when to sleep a service. For Railway to put a service to sleep, a service must not send outbound traffic for at least 10 minutes. Outbound traffic can include telemetry, NTP, etc.
- Enabling App Sleeping will apply the setting across all [Replicas](https://docs.railway.app/develop/services#horizontal-scaling-with-replicas)
```

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Scale to Zero
+---
+
```
half0wl, line 4: I'd add the Priority Boarding banner here, as it's a beta feature. E.g.:
ss 2023-10-05 at 11 37 00 PM

src/docs/reference/scale-to-zero.md
```diff
@@ -0,0 +1,49 @@
...
+App sleeping enables automated detection of an inactive [service](/develop/services) based on outbound traffic.
+
+Enabling App Sleep on a service tells Railway to stop the service when it is inactive, effectively reducing the overall cost to run it.
+
+<Image src="https://res.cloudinary.com/railway/image/upload/v1696017787/events_gjkaob.png"
+alt="App Sleeping in the Activity Feed"
+layout="intrinsic"
+width={700} height={460} quality={100} />
+
+### Inactive Service Detection
```
half0wl, line 18: Thoughts on making "inactive" / "active" into "slept" and "wake" for consistency? This feels like it may be semantically referring to the same thing using different terms

src/docs/reference/scale-to-zero.md
half0wl: Reasoning/questions:
- "this setting will be applied to replicas" have nothing to do with enabling this feature; it's a caveat, not a step in enabling it
- "an event will be visible in the dashboard" - need to be more specific, but would also throw this _outside_ of the steps to enable (intermingling instructional & informational content may cause confusion)

src/docs/reference/scale-to-zero.md
melissa-hale: It is truly only outbound traffic that can wake up a service.  We can clarify this if you think users will often get tripped up.  I'll leave it as-is for now, and we can figure out a good way to wordsmith it

@20k-ultra

src/docs/reference/scale-to-zero.md
20k-ultra: This is funny because we did use Caveats but switched it to be more neutral wording instead of something that sounds negative. I'm ok with either route.

src/data/sidebar.ts
melissa-hale: We talked about this, and you gave the vote we needed.  Changed to App Sleeping

src/docs/reference/app-sleeping.md
```diff
@@ -0,0 +1,48 @@
+---
+title: App Sleeping
+---
+
+<PriorityBoardingBanner />
+
+App Sleeping allows you to increase the efficiency of resource utilization on Railway.  It may reduce the resource usage cose of a [service](/develop/services) by ensuring you only use what you need, when you need it.
```
20k-ultra, line 7:
```suggestion
App Sleeping allows you to increase the efficiency of resource utilization on Railway.  It may reduce the resource usage cost of a [service](/develop/services) by ensuring you only use what you need, when you need it.
```

PR #368, opened 2023-11-04
Update volume size limits and add a note about growing a volume

brody192:
Updates the volume limits as described by [Changelog #0156](https://railway.app/changelog/2023-10-20-database-templates#fixes-and-improvements)
Also adds a small note on the ability to grow a volume.


melissa-hale:
Looks good, thank you @brody192 !

PR #373, opened 2023-11-09
Melissa/log 779 variables love

melissa-hale:
reworked a lot of the variables page with the intent to make it easier to understand overall maintenance, and particular attention to reference variables.

### DIFF
--- a/src/docs/develop/variables.md
+++ b/src/docs/develop/variables.md
@@ -34,6 +34,7 @@ alt="Screenshot of Raw Editor"
 layout="responsive"
 width={1954} height={1303} quality={100} />
 
+
 ## Shared Variables
 
 Shared variables help reduce duplication of variables across multiple services within the same project.  They are defined at the project environment level and can be added in Project Settings > Shared Variables.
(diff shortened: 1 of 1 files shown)


src/docs/develop/variables.md
```diff
@@ -2,29 +2,41 @@
...
+In Railway, there is also a notion of configuration variables which allow you to control the behavior of the platform.
+
+## Types of Variables
+
+There are four main types of variables in Railway - 
+- **Service-level variables** are configured in your services' settings.
+- **Shared variables** are configured in your projects' settings and can be made available to any service in specific project.
+- **Railway-provided variables** are set by Railway to provide metadata about the environment, project, or service.
+- **User-provided configuration variables** are a group of reserved variables that change the behavior of Railway when set.
+
```
JakeCooper, line 23: I feel a bit overwhelmed by this section being upfront

src/docs/develop/variables.md
```diff
@@ -33,13 +45,83 @@ alt="Screenshot of Raw Editor"
...
 width={1954} height={1303} quality={100} />
 
-Variables can be defined as simple key/value pairs or as [Templated Variables](#templated-variables) (eg. `${{Postgres.DATABASE_URL}}`),
-which can dynamically reference other variables, shared variables, or plugin variables (more on this below).
+
+#### Using Reference Variables
+
+Reference variables refer to variables set on a service that reference variables which are defined elsewhere, i.e. in another service or Shared variables.
+
+More on reference variables [below](/develop/variables#reference-variables).
```
JakeCooper, line 53: This section is introduced, then buried, then re-introduced later

Propose moving references up so that, as users evolve their application, they can go "deeper" into the docs

JakeCooper (commented):
Bones are excellent

Proposed a bit of re-ordering due to the way that I personally would think about "growing my usage" of variables on railway (single variables, references, shared using references, railway provided using references)

melissa-hale:
Thanks @JakeCooper ! super helpful feedback.  I think it's a lot better now

src/docs/develop/variables.md
melissa-hale: agreed, axed

src/docs/develop/variables.md
melissa-hale: moving things around made this flow perfectly

src/docs/develop/variables.md
```diff
@@ -2,27 +2,28 @@
...
 <Image src="https://res.cloudinary.com/railway/image/upload/c_scale,w_2026/v1678820924/docs/CleanShot_2023-03-14_at_12.07.44_2x_rpesxd.png"
 alt="Screenshot of Variables Pane"
 layout="responsive"
 width={2026} height={933} quality={100} />
 
-You can view all variables for the current environment using the [Railway CLI](/develop/cli) with
-`railway variables` and change the environment with `railway environment`.
+Click on `New Variable` to enter your new variable into a form field or use the `RAW Editor`.
 
 ### Raw Editor
```
JakeCooper, line 28: Now, reading through, is the raw editor required in the docs, or can we get by without it? I'm unsure if we need the pictures as it kinda "breaks up" the logical flow

src/docs/develop/variables.md
```diff
@@ -33,13 +34,92 @@ alt="Screenshot of Raw Editor"
...
+When using reference variables, you also have access to [Railway-provided variables](/develop/variables#railway-provided-variables).
+
+### Example Scenarios
 
-### Multiline Variables
+Here are some example scenarios to help clarify reference variable usage and syntax.
+
+#### Referencing a Shared variable -
+- You have a shared variable defined in your project called `API_KEY`, and you need to make the API key available to a service.  Go to the service's variables tab, and add a variable with the following value:
+  - `API_KEY=${{shared.API_KEY}}`
```
JakeCooper, line 70: So, since variable references are required for shared variables, should we introduce references before them?

Here, you have to double back again

src/docs/develop/variables.md
melissa-hale: I know what you mean, and I really think the restructure addresses the issue.  There should be a section that is mostly narrative - few to none pictures, no "this is how you do this" - and then we should have other pages (guides) where we put the images and helper information about how to do something and where you find things.

melissa-hale: for the sake of this PR though, I kinda think we should keep the image because it does help show people what we're talking about when we say "Raw Editor".  I'm okay with taking it out though if you think it's too distracting

JakeCooper: Love it. You're driving, so please just take my comments as random thoughts

Take em or leave em; totally fine either way!

src/docs/develop/variables.md
melissa-hale: Tried it different ways, I think I like Shared before Reference because you also see Shared in Reference so it's circular.  but this way, as soon as you get to the end of the Shared section that talks about how it creates a reference, you go right into Reference.

PR #405, opened 2023-11-28
Update templates.md

haydenlingle:
The hobby tier no longer has a $5 discount; it is $5/month with included usage. Line removed looks like it was added when there was the $5 discount awhile ago. 

It may be worth re-visiting logic in the billing system to make sure it's updated too.

### DIFF
--- a/src/docs/reference/templates.md
+++ b/src/docs/reference/templates.md
@@ -126,7 +126,6 @@ If your published template is deployed into other users' projects, you are immed
 #### Important things to note
 
 - Your template must be published to the marketplace to be eligible for kickback.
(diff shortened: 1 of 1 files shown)


brody192:
Some users have the fee waived, this line is still valid

haydenlingle:
> Some users have the subscription fee waived, this line is still valid

Thanks for clarifying!

ndneighbor:
Hey there @haydenlingle- this line is in reference to the kickback mechanics, we do offer the $5 discount for certain accounts with activity or company signups. I sincerely appreciate you opening up the PR, but will have to close this one out for now.

Issue #408, opened 2023-12-05
Compare to heroku page not found


finnala:
Hi, if you navigate to the "compare to Heroku" page on the docs, you get a page not found:

<img width="1920" alt="image" src="https://github.com/railwayapp/docs/assets/122728490/ad32bb6d-3492-4e81-a3f9-cc205d104d71">

Which is on route https://docs.railway.app/maturity/compare-to-heroku.

If you go from click the "continue reading" button under "what we offer over heroku" on the main heroku page (https://railway.app/heroku), it takes you to it.
<img width="1920" alt="image" src="https://github.com/railwayapp/docs/assets/122728490/b91fdb99-010d-4889-88c4-646ebad6d20e">
<img width="1920" alt="image" src="https://github.com/railwayapp/docs/assets/122728490/e15d6ef7-eb36-4fa6-803e-0e0de3d21392">

Looks like the `href` on the continue reading button takes you to https://docs.railway.app/reference/compare-to-heroku. 

But, you can't find this tab under "Reference" on docs, and it's instead under the "Maturity" tab which leads to a 404.

I think the ideal fix IMO is to fix the main docs link (put the compare to heroku docs under migrate), and then adjust the link on the https://railway.app/heroku page.

melissa-hale:
Hi @finnala, thank you for raising!  I have corrected this via a redirect.

PR #409, opened 2023-12-05
Add compare to heroku file to maturity

finnala:
In relation to #408 

Added the `compare to heroku` page under the `maturity` section in docs.

I haven't removed the file from `reference` yet, because the main website links to it from the continue reading button on https://railway.app/heroku. As the main website isn't open source (as far as I can tell) I thought it would be best to wait until this PR is merged, the main website link to be updated, before deleting the file in `reference`. I hope this makes sense.

Also this "works on my machine" but I had some issues with NPM, so I did this through GitHub on the web haha😅

 Hoping that it looks all good here 🙂

### DIFF
--- a/src/docs/maturity/compare-to-heroku.md
+++ b/src/docs/maturity/compare-to-heroku.md
@@ -1,0 +1,70 @@
+---
+title: Railway vs. Heroku
+---
+
+## Summary
+
+Railway is a simple and powerful deployment platform that focuses on giving you and your team a deployment plane that radically increases developer efficiency.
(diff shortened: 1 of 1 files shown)


melissa-hale:
Thank you very much @finnala !  I corrected this here: https://github.com/railwayapp/docs/pull/410/commits/04f795d9abef30b86e51c29534be9258ae93d031

I had to move the file, but I also setup a redirect so we don't break the link.  We'll go back and fix the links properly at some point.

PR #410, opened 2023-12-06
Melissa/log 1113 small ish content updates

melissa-hale:
lots of mostly small content updates based on feedback

### DIFF
--- a/src/docs/reference/public-networking.md
+++ b/src/docs/reference/public-networking.md
@@ -15,11 +15,11 @@ Railway can detect a deployed service is listening for traffic.  When detected, 
 | Category | Key Information |
 |----------|-----------------|
 | **DNS/Domain Names** | - Support for domains, subdomains, and wildcard domains.<br /> - Subdomains and wildcards cannot overlap (`foo.hello.com` cannot exist with `*.hello.com` unless owned by the same service).<br /> - Root domains need a DNS provider with ALIAS records or CNAME flattening.<br /> - Unicode domains should be PUNYcode encoded.<br /> - Non-public/internal domain names are not supported. |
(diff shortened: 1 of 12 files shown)


src/docs/reference/public-networking.md
```diff
@@ -0,0 +1,54 @@
...
+| **Certificate Issuance** | - Raiway attempts to issue a certificate for **up to 72 hours** after domain creation before failing.<br /> - Certificates are expected to be issued within an hour. |
+| **TLS** | - Support for TLS 1.2 and TLS 1.3 with specific ciphersets.<br /> - Certificates are valid for 90 days and renewed every 30 days.<br /> - Cloudflare Proxying does impact certificate issuance and support for wildcard domains. |
+| **Edge Traffic** | - Support for HTTP/1.1 and HTTP/2.<br /> - Support for websockets over HTTP/1.1 (may be interrupted after 2-4hrs, clients should handle reconnect). <br /> - Idle timeout of 900 seconds.<br /> - Max 100 request headers.<br /> - Max 100 concurrent streams per HTTP2 connection.<br /> - Max duration of 5 minutes for HTTP requests. |
+| **Request Headers** | - `X-Envoy-External-Address` or `X-Forwarded-For` for identifying client's remote IP.<br /> - `X-Forwarded-Proto` always indicates HTTPS.<br /> - `X-Request-Id` for correlating requests against network logs. |
```
char8, line 21: `X-Envoy-External-Address` - lets not mention this since it may change. `X-Forwarded-For` should be the same.

src/docs/reference/public-networking.md
```diff
@@ -0,0 +1,54 @@
...
+| **Requests** | - Inbound traffic must be TLS-encrypted <br /> - HTTP GET requests to port 80 are redirected to HTTPS. <br /> - HTTP POST requests to port 80 are redirected to HTTPS as GET requests.|
```
char8, line 22: shall we also add `SNI is required for correct certificate matching`

PR #470, opened 2024-04-09
Remove all references of plugins

Milo123459:


zuchka:
can we also update this file?

https://github.com/railwayapp/docs/blob/main/src/docs/reference/cli-api.md

zuchka (approved):
GREAT! thank you!

PR #492, opened 2024-05-31
Beta Observability Dashboard Documentation

ndneighbor:
Title says it all!


src/data/sidebar.ts
melissa-hale, line 1: can we put this under Guides/Monitoring instead?

melissa-hale:
Looks great!!  since it's more how-to-y, I think it should go under Guides, otherwise ✅

src/data/sidebar.ts
ndneighbor: Done! Yea, might have started in once place and turned out the other way.

PR #494, opened 2024-06-03
Update public-networking.md custom domain section

tambuildsthings:
Adding a line to Important Considerations to explain that Trial plan users only have one custom domain available.  
It currently isn't clear that www and root domains are considered unique in Railway.  
There are [several posts in the Discussions](https://help.railway.app/search?q=www+redirect) and [Discord](https://discord.com/channels/713503345364697088/1100135019726508164) asking about this so I'm trying to clarify directly in the docs.


half0wl (approved):
Great callout, thank you!

PR #518, opened 2024-07-25
docs: should define absolute path for railway config file

mittalyashu:


mittalyashu:
//cc @FarazPatankar

melissa-hale:
Thanks @mittalyashu !

PR #520, opened 2024-07-25
Add Cloudflare Proximity Steering Tutorial

brody192:
^

### DIFF
--- a/src/docs/tutorials/proximity-steering.md
+++ b/src/docs/tutorials/proximity-steering.md
@@ -78,7 +78,7 @@ width={1060} height={555} quality={100} />
 
 - Leave Port 443 as the default.
 
-<Image src="https://res.cloudinary.com/railway/image/upload/v1721867906/docs/tutorials/proximity-load-balancing/pool_settings_health_ydlzvo.png"
+<Image src="https://res.cloudinary.com/railway/image/upload/v1722015787/docs/tutorials/proximity-load-balancing/health_monitor_oty6pd.png"
 alt="screenshot of the cloudflare health monitor"
 layout="responsive"
 width={1060} height={375} quality={100} />
--- a/src/docs/tutorials/proximity-steering.md
+++ b/src/docs/tutorials/proximity-steering.md
@@ -99,7 +99,7 @@ width={1060} height={375} quality={100} />
 
     Example -  `region-us-west1.up.railway.app`
 
-    This should be just the domain without the scheme or trailing slash.
+    This should only be the domain, excluding both the scheme and trailing slash.
 
 - For the weight option we will use a value of **1**.
 
--- a/src/docs/tutorials/proximity-steering.md
+++ b/src/docs/tutorials/proximity-steering.md
@@ -1,8 +1,8 @@
 ---
-title: Setting up Proximity Load Balancing using Cloudflare
+title: Setting up Proximity Steering using Cloudflare
 ---
 
-### What is Proximity Load Balancing?
+### What is Proximity Steering?
 
 >Proximity steering routes visitors ... to the closest physical data center.
 
@@ -10,9 +10,9 @@ title: Setting up Proximity Load Balancing using Cloudflare
 
 ## About this Tutorial
 
-As Railway does not yet offer native Geo or Proximity Load Balancing we instead need to place Cloudflare in front of our services to do this for us.
+As Railway does not yet offer native Proximity Steering we instead need to place Cloudflare in front of our services to do this for us.
 
-This tutorial aims to provide a simple step-by-step on how to set up everything on Cloudflare in order to have Proximity Load Balancing working flawlessly!
+This tutorial aims to provide a simple step-by-step on how to set up everything on Cloudflare in order to have Proximity Steering working flawlessly!
 
 **Objectives**
 
@@ -26,20 +26,26 @@ In this tutorial, you will learn how to -
 
 - Have two or more identical services deployed in two or more different regions.
 
+    Duplicating a service can be done by right clicking and selecting **Duplicate**, opening its service settings and changing the region, then clicking **Deploy**.
+
+    The services you have deployed into the multiple regions **should only** have a Railway generated domain on them, it also helps to have the domains indicate the region.
+
+    Its recommended to use [shared variables](/guides/variables#shared-variables) or [reference variables](/guides/variables#referencing-another-services-variable) for duplicated services to keep variables in sync.
+
 <Image src="https://res.cloudinary.com/railway/image/upload/v1721867908/docs/tutorials/proximity-load-balancing/region_services_u10ukp.png"
 alt="screenshot of two railway services in different regions"
 layout="responsive"
 width={890} height={435} quality={100} />
 
-    Duplicate a service to another region is as simple as right clicking and selecting **Duplicate**, opening its service settings and changing the region, then clicking **Deploy**.
-
-    The services you have deployed into the multiple regions **should only** have a Railway generated domain on them, it also helps to have the domains indicate the region.
-
 - Have your desired domain setup with Cloudflare's nameservers, they have a general guide for that [here](https://developers.cloudflare.com/dns/zone-setups/full-setup/setup/).
 
-- Have **SSL/TLS** mode set to **Full** - **SSL/TLS → Overview → Full**
+- Have **SSL/TLS** mode set to **Full**
 
-- Have **Always Use HTTPS** enabled - **SSL/TLS → Edge Certificates → Always Use HTTPS**
+    **SSL/TLS → Overview → Full**
+
+- Have **Always Use HTTPS** enabled
+    
+    **SSL/TLS → Edge Certificates → Always Use HTTPS**
     
     This is to prevent Railway from needing to handle the insecure redirect as that would result in being redirected away to an upstream endpoint.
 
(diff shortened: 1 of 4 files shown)


src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+
+## About this Tutorial
+
+As Railway does not offer native Proximity Steering at this time, we instead need to place Cloudflare in front of our services to do this for us.
+
+This tutorial aims to provide a simple step-by-step guide on setting everything up on Cloudflare to ensure Proximity Steering works flawlessly!
+
+**Objectives**
+
+In this tutorial, you will learn how to -
```
melissa-hale, line 19: I would clarify that each of these objectives refer to actions that will be performed in CloudFlare.  e.g.. "... you will learn how to do the following in CloudFlare - "

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Create a Health monitor.
+- Create pools for each region.
+- Set up the Proximity Load Balancer.
+
+**Prerequisites**
+
+- Have two or more identical services deployed in two or more different regions.
```
melissa-hale, line 27: similar here, clarify that we are now in the context of Railway.  "Have two or more identical services .... in Railway"

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+    Duplicating a service can be done by right clicking and selecting **Duplicate**, opening its service settings and changing the region, then clicking **Deploy**.
+
+    The services you have deployed into the multiple regions should **only** have a Railway generated domain on them, it also helps to have the domains indicate the region.
```
melissa-hale, line 32: maybe - "The services should be configured with a Railway-generated domain, do not assign a custom domain.  It is also helpful to indicate the region in the domain."

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+
+    Its recommended to use [shared variables](/guides/variables#shared-variables) or [reference variables](/guides/variables#referencing-another-services-variable) for duplicated services to keep variables in sync.
```
melissa-hale, line 34: *It is or It's

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+**Objectives**
+
+In this tutorial, you will learn how to -
+
+- Create a Health monitor.
+- Create pools for each region.
+- Set up the Proximity Load Balancer.
+
+**Prerequisites**
+
```
melissa-hale, line 26: Maybe add a line here that's just.

"In Railway -"

and then below here when you get into CF settings, we can do the same... i'll make a comment there too

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+    The services you have deployed into the multiple regions should **only** have a Railway generated domain on them, it also helps to have the domains indicate the region.
+
+    Its recommended to use [shared variables](/guides/variables#shared-variables) or [reference variables](/guides/variables#referencing-another-services-variable) for duplicated services to keep variables in sync.
+
+<Image src="https://res.cloudinary.com/railway/image/upload/v1722015743/docs/tutorials/proximity-load-balancing/region_services_u10ukp.png"
+alt="screenshot of two railway services in different regions"
+layout="responsive"
+width={890} height={435} quality={100} />
+
```
melissa-hale, line 40: Add
"In CloudFlare - "

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+    **SSL/TLS → Overview → Full**
+
+- Have **Always Use HTTPS** enabled.
+    
+    **SSL/TLS → Edge Certificates → Always Use HTTPS**
+    
+    This ensures that Railway avoids managing the insecure redirect, which would otherwise lead to an incorrect redirection to an upstream endpoint.
+
+- Have a `/health` or similar endpoint that returns a status code of 200.
```
melissa-hale, line 53: "Have a `/health` or similar endpoint .... in the services deployed to Railway"

src/docs/tutorials/proximity-steering.md
brody192: that flows much better!

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+
+    This allows Cloudflare to check the health of our Railway services so they can handle region failover. As a bonus this can also be used on Railway to achieve [zero-downtime deployments](/reference/healthchecks).
```
melissa-hale, line 55: love this mention!

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+
+
```
melissa-hale, line 57: Same thought here.. clarify that all of the steps below will be done inside of CloudFlare

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+- Click **Save**.
+
+## 2. Creating the Pools
+
+- Go back to the Load Balancing page.
+
+- Click **Manage Pools** and then **Create**.
+
+- Fill out the name and description and leave **Endpoint Steering** as its default, it will not be used with only a single endpoint.
```
melissa-hale, line 94: lil confusing.  in the screenshot, i see "Random" for this field.  Is that the default?  if so, maybe smth like ".. leave Endpoint Steering as the default of 'Random' .. "

src/docs/tutorials/proximity-steering.md
brody192: I thought you would! :)

src/docs/tutorials/proximity-steering.md
brody192: Yeah, random is the default, I will clarify

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+
+    Example -  `region-us-west1.up.railway.app`
+
+    This should only be the domain, excluding both the scheme and trailing slash.
+
+- For the weight option we will use a value of **1**.
+
+- Click **Add host header** and enter the same value as used for the Endpoint Address.
+
+    Railway does host based routing so we need to set the host header so they know how to route the incoming requests from Cloudflare.
```
melissa-hale, line 108: lil confusing because I feel like I need to take an action after I read this line, but it's really just informational.  maybe clarify like "this step is important since Railway uses host-based routing and requires the host header to know how to route the incoming requests from Cloudflare"

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+- Remove the second empty endpoint.
+
+    Our pool only needs to contain a single endpoint as Railway handles single region replicas for us.
+
+<Image src="https://res.cloudinary.com/railway/image/upload/v1722015878/docs/tutorials/proximity-load-balancing/pool_settings_config_qh5s1k.png"
+alt="screenshot of end endpoint settings in the pool creator"
+layout="responsive"
+width={1060} height={600} quality={100} />
+
+- Click **Configure coordinates for Proximity Steering** and enter the Latitude and Longitude for your service region that can be found in this [JSON file](https://www.google.com/about/datacenters/json/locations.json).
```
melissa-hale, line 119: how about adding an example of how to do this?  i opened that json file, and assumed i was looking for like us-west but don't see that, so confused

src/docs/tutorials/proximity-steering.md
```diff
@@ -0,0 +1,203 @@
...
+layout="responsive"
+width={1060} height={375} quality={100} />
+
+- Click **Save**.
+
+- Create another pool for your other services that are deployed into your desired regions, follow the same procedure.
+
+This should be the end result, two or more pools -
+
+<Image src="https://res.cloudinary.com/railway/image/upload/v1722015821/docs/tutorials/proximity-load-balancing/pools_w1gext.png"
```
melissa-hale, line 141: 🥳

melissa-hale:
This is great!!!  I think it would be good to mention somewhere, maybe at the top when you define proximity steering, to also mention aka "geo-load balancing" or whatever other term(s) you see people use.  I don't hear "proximity steering" often, so we wanna make sure if people search, they find this tutorial

and last thought i promise, how would you feel about adding like an "Additional resources" section and providing links to relevant CloudFlare docs?  Like, I assume the load balancer can be configured in other ways, so maybe just acknowledge that and point ppl to the CF docs to learn more

src/docs/tutorials/proximity-steering.md
brody192: that's fair, I will add that

brody192:
I actually started with the title being geo load balancing but then as I was looking at cloudflare's docs they called it proximity steering, so I did to, but I will add those extra keywords as well!

I will also definitely add an additional resources section with various relevant cloudflare docs.

src/docs/tutorials/proximity-steering.md
brody192: I agree, will change!

brody192:
I think I have implemented all the suggestions!

melissa-hale:
Merging!

PR #522, opened 2024-07-26
[PRO-2658] Document private registry credentials

evancloutier:
## Why

Documents how to use private registry credentials in the product 👍


zuchka (approved):
lgtm!

PR #547, opened 2024-08-25
Update quick-start.md

ritikaaaa177:
Steps should always be explained in numbers. So, I modified them.


brody192:
Thank you for the contribution but I prefer these as bullet points.

We hope to see you continue to contribute in the future!

PR #551, opened 2024-09-05
Add Best Practices

brody192:
https://linear.app/railway/issue/LOG-2195/[docs]-best-practices

### DIFF
--- a/src/docs/overview/best-practices.md
+++ b/src/docs/overview/best-practices.md
@@ -1,0 +1,30 @@
+---
+title: Best Practices
+---
+
+Railway is a highly versatile platform, offering various ways to use it, though some may be less optimal than others for most use cases. the topics here aim to help you maximize the platform's potential.
+
+## Use Private networking
+
+Private networking allows services within a project to communicate internally without the need to expose them publicly.
+
+The private network has no egress fees, allowing you to avoid service-to-service egress costs when communicating with a database or other services within the same project.
+
+## Deploying related Services into the same Project
+
+It can be all too easy to create a new project each time you deploy a database or service, but there are a few reasons to keep related services within the same project.
+
+- **Private networking** - The private network is scoped to a single environment within a project, having all related services within a single project will allow you to use private networking for faster networking along with no egress fees for service to service communication.
+
+- **Project clutter** - Deploying a new service or database as an entire project will quickly become overwhelming and clutter your dashboard.
+
+## Use Reference variables
+
+Reference variables allow you to reference to another variable, either within the current service or from another service in the same project.
+
+This is recommended over copying and pasting variables you need to access in another service, for example.
+
+Using reference variables ensures your variable values are always in sync. Changed your TCP proxy? The variable updates. Changed your backend hostname? The variable updates.
+
+For examples on how to use reference variables, please see our [reference variable guide](/guides/variables#reference-variables).
+
--- a/src/docs/overview/best-practices.md
+++ b/src/docs/overview/best-practices.md
@@ -1,25 +1,44 @@
 ---
 title: Best Practices
 ---
+{/**
+To keep constancy we want each topic to follow the same convention -
+- What?
+- When?
+- Do X.
+- Why?
+- Image.
+*/}
 
-Railway is a highly versatile platform, offering various ways to use it, though some may be less optimal than others for most use cases. the topics here aim to help you maximize the platform's potential.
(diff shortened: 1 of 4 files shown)


src/docs/overview/best-practices.md
```diff
@@ -6,25 +6,24 @@ Railway is a highly versatile platform, offering various ways to use it, though
 
 ## Use Private networking
 
-Private networking allows services within a project to communicate internally without the need to expose them publicly.
+[Private networking](/reference/private-networking) allows services within a project to communicate internally without the need to expose them publicly.
 
 The private network has no egress fees, allowing you to avoid service-to-service egress costs when communicating with a database or other services within the same project.
```
melissa-hale, line 12: Also mention that communication will be much faster too

src/docs/overview/best-practices.md
```diff
@@ -0,0 +1,29 @@
...
+Railway is a highly versatile platform, offering various ways to use it, though some may be less optimal than others for most use cases. the topics here aim to help you maximize the platform's potential.
+
+## Use Private networking
+
+[Private networking](/reference/private-networking) allows services within a project to communicate internally without the need to expose them publicly.
+
+The private network has no egress fees, allowing you to avoid service-to-service egress costs when communicating with a database or other services within the same project.
+
+## Deploying related Services into the same Project
+
```
melissa-hale, line 14: maybe add some words about how reference variables make your variable assignments dynamic.  instead of hardcoding your variable values, use reference variables so that they update when the value changes.

src/docs/overview/best-practices.md
```diff
@@ -0,0 +1,29 @@
...
+It can be all too easy to create a new project each time you deploy a database or service, but there are a few reasons to keep related services within the same project.
+
+- **Private networking** - The private network is scoped to a single environment within a project, having all related services within a single project will allow you to use private networking for faster networking along with no egress fees for service-to-service communication.
+
+- **Project clutter** - Deploying a new service or database as an entire project will quickly become overwhelming and clutter your dashboard.
+
+## Use Reference variables
+
```
melissa-hale, line 22: maybe add some words about how reference variables make your variable assignments dynamic. instead of hardcoding your variable values, use reference variables so that they update when the value changes.

melissa-hale:
Looks great! and we can keep adding stuff to it.  I think what's missing for me, are some graphics.  Given that this page is at the top in the overview section, people may read this before learning what any of this stuff is.  The links are good, but some pictures might be easier on the brain.

src/docs/overview/best-practices.md
brody192: I have, just a few more lines down 😄

> Using reference variables ensures your variable values are always in sync. Change your TCP proxy? The variable updates. Change your backend hostname? The variable updates.

brody192:
Screenshots are a great idea, I'll try to get some that best represent via the UI the given topic.

brody192:
Screenshots implemented, with little descriptions!

src/docs/overview/best-practices.md
melissa-hale: Ah okay hmm .. I feel like we can make things more clear..  Now that I think about it, this applies for each section actually.

I think you should rework them a bit so they all follow a similar pattern.. start off with the scenario (_when doing X_) then the directive (_do Y_) then state why (_because Z_), then show what it looks like and point to the appropriate docs page.  e.g. 

"When connecting your services in Railway, always use the private network.  Communication is much faster and you will avoid egress fees" ... then your graphic and link to the private network docs.

whatcha think?

brody192: Yeah that makes sense, might be better than just being told to do something!

melissa-hale:
Oops somehow missed the screenshot comment, they look great!!  

Take a look at my last comment above and let me know what you think.  I think if you do that, it'll sing.  should mostly just be a matter of moving stuff around and clarifying when these activities become relevant (connecting services, adding new services, adding variables to services)

src/docs/overview/best-practices.md
brody192: I think it's far easier to read now that each topic follows the same pattern!

melissa-hale: Looks great @brody192 ❤️  (sorry if you were waiting on me)

melissa-hale (approved):

PR #553, opened 2024-09-11
add docs for sealed variables

FarazPatankar:
^

Add docs for sealed variables.

### DIFF
--- a/src/docs/guides/variables.md
+++ b/src/docs/guides/variables.md
@@ -107,18 +107,38 @@ width={2408} height={1150} quality={100} />
 
 ## Sealed Variables
 
-Sealed variables are useful when you want to provide a value to your builds and deployments but not have it be visible in the UI (or via the API).
+<PriorityBoardingBanner />
 
-### Sealing a variable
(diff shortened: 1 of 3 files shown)


src/docs/guides/variables.md
```diff
@@ -96,9 +105,29 @@ alt="Screenshot of Variables Pane"
 layout="responsive"
 width={2408} height={1150} quality={100} />
 
+## Sealed Variables
+
+Sealed variables are useful when you want to provide a value to your builds and deployments but not have it be visible in the UI (or via the API).
```
melissa-hale, line 110: How about something like - 

Railway provides the ability to **seal** variable values for extra security.  When a variable is sealed, its value is provided to builds and deployments but is never visible in the UI nor can it be retrieved via the API.

src/docs/guides/variables.md
```diff
@@ -96,9 +105,29 @@ alt="Screenshot of Variables Pane"
...
+
+### Sealing a variable
+
+To seal a variable, click the 3-dot menu on the right-side of an existing variable and click the `Seal` button.
+
+### Caveats
```
melissa-hale, line 116: mention they cannot be updated via the Raw Editor?  or like Brody mentioned, add something about how to edit them.  could be after "Sealing a variable", "Updating a sealed variable"

PR #558, opened 2024-09-17
Improvements!

brody192:
### This PR fixes some of my long-standing and newfound gripes -

- Inline code blocks were just colored text with backticks, we now render inline code blocks similarly to how they are rendered everywhere else, with a background and without backticks.

- On some browsers, jump links would not scroll to the correct location on pages with code blocks, this was because code blocks were being forced to render client side causing the final scroll location to have an offset after the code blocks were rendered.

- The page navigation menu snapped the scroll to header elements instead of smooth scrolling to them, similarly the same happened when clicking the header elements themselves, instead of them being hrefs they now have a click handler to scroll to the element's id.

- Collapse elements could not be used as a jump link like headers can, we now store the last opened collapse as a hash in the url, and upon page load, open the corresponding collapse element and jump to it.

- Collapse elements used the default bullet, we now use the same arrow as the sidebar does.

- You couldn't easily jump link to h4 elements without the page navigation menu, we now update the url's hash when you click an h4 element and that will have the browser scroll to it on load.

- All links to sections on the same page will trigger a smooth scroll when clicked.

### This PR also adds some additional information / fixes -

- Update list of DNS providers that support root level CNAMEs - [Bunny.net](http://bunny.net/)

- Update public and private networking docs to mention the need for dual-stack binding then applications need to be accessible over both public and private networks.

- Adds info about what you can expect when you use target ports without listening on or setting the `PORT` environment variable.

- Add a new "Deployment Menu" section that explains what each action in a deploy's 3-dot menu does.

- Adds a new FAQ selection that explains why deployments may be removed and the solutions, super common question.

- Add missing GitLab Container Registry in some spots that talked about docker hub, ghcr, and quay.

- Use an X icon instead of Twitter, which matches the chosen color theme.

- Add a little more information to the states of deployments.

- Update deprecated `images.domains` to `images.remotePatterns` in `next.conf.js`


next.config.js
```diff
@@ -1,15 +1,15 @@
 const { withContentlayer } = require("next-contentlayer");
-const redirects = require("./redirects"); 
+const redirects = require("./redirects");
```
ndneighbor, line 2: Not needed for this pr, but I have been convinced that alias paths are the way forward.

src/components/Header.tsx
```diff
@@ -0,0 +1,82 @@
...
+    ${StyledLinkIcon} {
+      display: flex;
+    }
+    @media (max-width: 1300px) {
+      ${StyledLinkIcon} {
+        display: none;
+      }
+    }
+  }
+`;
```
ndneighbor, line 45: Curious here, is it not possible to use something like Tailwind CSS?

src/components/InlineCode.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { PropsWithChildren } from 'react';
+import tw from 'twin.macro';
```
ndneighbor, line 2: Oh dear, okay, this is why, we may need to finally bite the bullet and port everything over to a more modern styling system.

ndneighbor (approved):
I realized my css comments are all tied to the status quo of the repo. I think this is great and massively improves on the copy we have so far.

next.config.js
brody192: Yeah they're used in some places but not everywhere.

src/components/Header.tsx
brody192: Definitely, but I didn't wanna touch this as it works and I don't know tailwind, I can come back to stuff like this in the future when I'm more familiar.

src/docs/reference/deployments.md
```diff
@@ -54,7 +58,57 @@ A Deployment will remain in the `Active` state unless it [crashes](/guides/deplo
...
+
+#### View logs
+
+Opens the deployment up to the corresponding logs, during build the build logs will be shown, during deploy the deploy logs will be shown.
+
+#### Restart
+
+Restarts the currently running process in the deployment, this is often used to bring a service back online after a crash or if you application has locked up.
+
+#### Redeploy
```
melissa-hale, line 87: selfish, but can you add a note about how we only retain images for 2 weeks. i.e. you can only roll back to a deployment that is 2 weeks old or newer (ref [LOG-2359](https://linear.app/railway/issue/LOG-2359/update-docs-to-include-2-week-retention-period-for-container-images))

src/docs/reference/deployments.md
```diff
@@ -54,7 +58,57 @@ A Deployment will remain in the `Active` state unless it [crashes](/guides/deplo
...
+- Restart a service after a usage limit has been reached.
+
+- Restart a service after upgrading to Hobby when trial credits are depleted.
+
+- Restart a service if you were demoted from Hobby to trial.
+
+**Note:** The redeploy will use the source code from the currently selected deployment.
+
+#### Rollback
+
```
melissa-hale, line 104: same here, mention the 2 week retention

src/docs/reference/deployments.md
```diff
@@ -42,9 +42,13 @@ If an error occurs during the build or deploy process, the Deployment will stop
 
 #### Active
 
-Once the Deployment is live and running, the status will change to `Active`.
+Railway will determine the deployment's active state with the following logic -
 
-### Completed
+- If the deployment **has** a healthcheck configured, Railway will mark the deployment as `Active` when the healthcheck succeeds.
```
melissa-hale, line 47: make 'healthcheck' a link to the reference page for healthchecks?

src/docs/guides/healthchecks-and-restarts.md
```diff
@@ -16,10 +16,14 @@ To configure a healthcheck -
 
 ## Configure Healthcheck Port
```
melissa-hale, line 17: i think this section needs an example or a screenshot

brody192: IMO a screenshot would be a tad overkill when all the user needs to do is set a service variable, I'll see what I can do to make that more straightforward.

brody192: Okay you've convinced me, I'll see what I can do for a screenshot that doesn't take up half the page.

melissa-hale: doesn't have to be crazy, can even come back to it, it's minor

brody192:
The old style of inline code blocks is what I've always hated the most (broken jump links are #2) I've never seen any markdown rendering show them like that, notion, GitHub, discord, etc, and now our docs render them in the same style.

I thought about adding the same color back to the text, but when I did that it was irritating to read for me.

But I'll throw in up in progress pic because now you've made me curious what other will think.

melissa-hale:
for sure, i totally get it, and I don't think anyone is going to push back.  the visual change is kinda big though, so just want to make sure it wasn't intentional that we had it like that for so long.  I'm really thinking of JR, so maybe we can just tag him here and get his stamp

melissa-hale:
BTW i did test out the scrolling changes and it's soo much better, great job 👏  all of the copy looks great too

brody192:
True, most of this code was @coffee-cup's

How do you feel about rendering inline code like [this](https://docs-frontend-docs-pr-558.up.railway.app/guides/public-networking#adding-a-root-domain-with-www-subdomain-to-cloudflare)?

src/hooks/useScrollToOpenCollapse.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect } from 'react';
+
+export function useScrollToOpenCollapse() {
```
coffee-cup, line 3: I looks like this file is 4 space per ident instead of 2. We should run prettier on everything to ensure consistent formatting

src/utils/scroll.ts
```diff
@@ -0,0 +1,17 @@
+export const scrollToID = (input: string, skipScroll: boolean = false) => (event: React.MouseEvent) => {
+    const id = input.split('#')[1] ?? input;
```
coffee-cup, line 2: This will strip a leading hash if it exists
```suggestion
    const id = id.replace(/^#/, '');
```

coffee-cup (approved):
Some files look to be 4 space instead of 2, but besides that the React looks sane

PR #562, opened 2024-09-19
Fix the color flashing on first page draw on code blocks

brody192:
- Fixes dark mode flash on code blocks when set to light mode
- Fixes the [X.com](http://x.com/) icon not respecting the set theme.
- Adds another section to the health check guide that mentions the use of `[healthcheck.railway.app](http://healthcheck.railway.app/)` as the hostname when health checks are performed.
- Changes some grammar about healthchecks not being used for continuous testing.
- Updates the copy for the fixing common errors page to add if clauses on the target port information. (some deploys may not be using them)
- Adds an FAQ section to the pricing faq page that covers how a user can go from a pre-paid hobby plan to an automatic payment hobby plan, or vice versa.

### DIFF
--- a/src/components/CodeBlock/index.tsx
+++ b/src/components/CodeBlock/index.tsx
@@ -39,11 +39,17 @@ export interface Props {
   language?: string;
   className?: string;
   children?: any;
+  colorModeSSR?: string | null;
 }
 
 const getParams = (
   className = "",
-): { language?: string; for?: string; always?: boolean } => {
+): {
+  language?: string;
+  languageFull?: string;
+  for?: string;
+  always?: boolean;
+} => {
   const [language, params = ""] = className.split(":");
 
   const splitParams = params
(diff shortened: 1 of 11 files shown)


src/components/CodeBlock/index.tsx
```diff
@@ -59,6 +65,7 @@ const getParams = (
 
   return {
     language: language.split("language-")[1],
+    languageFull: language || "language-plaintext",
```
coffee-cup, line 68: Is this used anywhere?

coffee-cup (approved):
This is certainly some jank. But jank that should hopefully fix the theme flash in the meantime :)

src/components/CodeBlock/index.tsx
brody192: nope! just a left over artifact from something i was testing.

PR #569, opened 2024-09-27
Languages & Frameworks: Add FastAPI guide (Python)

unicodeveloper:
Completes Linear Issue [#LOG-2573](https://linear.app/railway/issue/LOG-2573).

CleanShot 2024-09-27 at 13 52 12@2x

### DIFF
--- a/src/data/sidebar.ts
+++ b/src/data/sidebar.ts
@@ -45,6 +45,7 @@ export const sidebarContent: ISidebarContent = [
         subTitle: makePage("Languages & Frameworks", "guides"),
         pages: [
           makePage("Fastify", "guides"),
+          makePage("FastAPI", "guides"),
         ],
       },
       {
--- a/src/docs/guides/fastapi.md
+++ b/src/docs/guides/fastapi.md
@@ -86,4 +86,5 @@ Explore these resources to learn how you can maximize your experience with Railw
 
 - [Add a Database Service](/guides/build-a-database-service)
 - [Monitor your app](/guides/monitoring)
+- [Running a Cron Job](guides/cron-jobs)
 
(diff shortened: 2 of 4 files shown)


brody192:
Guides 2, 3, and 4 all talk about deploying the template's repo, what if I want to deploy my own app?

Apologies for not thinking about this point sooner given the previous guide doesn't cover this topic either.

unicodeveloper:
> Guides 2, 3, and 4 all talk about deploying the template's repo, what if I want to deploy my own app?
> 
> Apologies for not thinking about this point sooner given the previous guide doesn't cover this topic either.

Not a problem. 

This is the reason I added both the Dockerfile and CLI option for folks that want to deploy their own apps because the readers (beginners) can switch to their own apps after trying the template's repo to theirs. And the not-beginners can just deploy their own apps with that knowlege.

What I was thinking for something more comprehensive is to write about how to deploy a full app with 4 services (Web, Cron, Worker, Database). But I plan to do that starting with the **PHP/Laravel** & **Ruby on Rails** guide.

Do you have another opinion? @brody192

brody192:
> the readers (beginners) can switch to their own apps after trying the template's repo

It might sound harsh, but don't over estimate the user's intuition, this guide (and the last one) doesn't mention they can deploy their own app so they might come out of the guide thinking they need to deploy the template's repo to start off even if they already written their code.

True story, that has happened.

I'm not suggesting to add another section, just change / add some wording to mention that users can bring their own repos.

unicodeveloper:
> > the readers (beginners) can switch to their own apps after trying the template's repo
> 
> It might sound harsh, but don't over estimate the user's intuition, this guide (and the last one) doesn't mention they can deploy their own app so they might come out of the guide thinking they need to deploy the template's repo to start off even if they already written their code.
> 
> True story, that has happened.
> 
> I'm not suggesting to add another section, just change / add some wording to mention that users can bring their own repos.

Wow. That's very interesting. I'll add a paragraph to mention that.

unicodeveloper:
@brody192 Made required changes across both guides to make it clearer.

brody192:
Yep, it's definitely clearer now!

unicodeveloper:
> Yep, it's definitely clearer now!

Awesome!!! Time to merge 🚀

PR #571, opened 2024-10-02
Add more info to app sleeping

brody192:
- Adds more information to the App Sleeping reference page. Specially, adding scenarios that will keep the app from sleeping and make it clearer that private network traffic will prevent the app from sleeping,
- Fix a smooth scroll bug where links to different pages would smooth scroll to an element on the current page if the link had a hash for an element that existed on the current page, we now check if the link's path is the same as the windows path before deciding to smooth scroll or treat it as a regular link.


brody192:
LGTM!

PR #573, opened 2024-10-03
Languages & Frameworks: Add Laravel guide (PHP)

unicodeveloper:
Completes Linear Issue [#LOG-2575](https://linear.app/railway/issue/LOG-2575).

CleanShot 2024-10-03 at 01 07 43@2x

CleanShot 2024-10-03 at 01 07 55@2x

### DIFF
--- a/src/docs/guides/laravel.md
+++ b/src/docs/guides/laravel.md
@@ -18,8 +18,6 @@ This template sets up a basic Laravel application along with a Postgres database
 
 We highly recommend that [you eject from the template after deployment](/guides/deploy#eject-from-template-repository) to create a copy of the repo on your GitHub account.
 
-**Note:** If you see a "500 Server Error" on your deployed URL, it’s likely because Laravel is missing the **APP_KEY** or other required environment variables. To fix this, go to the Railway dashboard and add the necessary environment variables—especially the **APP_KEY**—and then redeploy the application. You can generate a new **APP_KEY** using this [Laravel APP KEY generator](https://generate-random.org/laravel-key-generator).
-
 ## Deploy from a GitHub Repo
 
 To deploy a Laravel app on GitHub to Railway, follow the steps below:
--- a/src/docs/guides/laravel.md
+++ b/src/docs/guides/laravel.md
@@ -21,22 +21,21 @@ We highly recommend that [you eject from the template after deployment](/guides/
 
 **Note:** If you see a "500 Server Error" on your deployed URL, it’s likely because Laravel is missing the **APP_KEY** or other required environment variables. To fix this, go to the Railway dashboard and add the necessary environment variables—especially the **APP_KEY**—and then redeploy the application. You can generate a new **APP_KEY** using this [Laravel APP KEY generator](https://generate-random.org/laravel-key-generator).
 
-
 ## Deploy from a GitHub Repo
 
-To deploy a Laravel app on Railway directly from GitHub, follow the steps below:
+To deploy a Laravel app on GitHub to Railway, follow the steps below:
 
-1. Fork the basic <a href="https://github.com/railwayapp-templates/laravel" target="_blank">Laravel GitHub repo</a>. 
-    - If you already have a GitHub repo you want to deploy, you can skip this step.
-2. Create a <a href="https://railway.app/new" target="_blank">New Project.</a>
-3. Click **Deploy from GitHub repo**.
-4. Select the `laravel` or your own GitHub repo.
+1. Create a <a href="https://railway.app/new" target="_blank">New Project.</a>
+2. Click **Deploy from GitHub repo**.
+3. Select your GitHub repo.
     - Railway requires a valid GitHub account to be linked. If your Railway account isn't associated with one, you will be prompted to link it.
-5. Click **Add Variables**. 
+4. Click **Add Variables**. 
     - Add all your app environment variables.
 5. Click **Deploy**.
 
 Once the deployment is successful, a Railway [service](/guides/services) will be created for you. By default, this service will not be publicly accessible.
+
+**Note:** Railway will automatically detect that it's a Laravel app during [deploy and run your app via php-fpm and nginx](https://github.com/railwayapp/nixpacks/tree/main/src/providers/php). 
 
 To set up a publicly accessible URL for the service, navigate to the **Networking** section in the [Settings](/overview/the-basics#service-settings) tab of your new service and click on [Generate Domain](/guides/public-networking#railway-provided-domain).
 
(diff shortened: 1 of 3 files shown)


coffee-cup (commented):
The formatting here is a bit strange
image

src/docs/guides/laravel.md
```diff
@@ -0,0 +1,185 @@
...
+            # Runs the Laravel scheduler every minute
+            while [ true ]
+                do
+                    echo "Running the scheduler..."
+                    php artisan schedule:run --verbose --no-interaction &
+                    sleep 60
+                done
+        ```
+2. Create a Database service on the <a href="/overview/the-basics#project--project-canvas" target="_blank">Project Canvas.</a>
+     - Click on **Deploy**.
```
coffee-cup, line 123: What type fo database service? I think this is a bit unclear

src/docs/guides/laravel.md
```diff
@@ -0,0 +1,185 @@
...
+- **Worker Service**: Processes background jobs from the queue.
+- **Database Service**: Stores and retrieves your application's data.
+
+<Image src="https://res.cloudinary.com/railway/image/upload/f_auto,q_auto/v1727910244/docs/quick-start/deploy%20architecture.png"
+alt="screenshot of the deploy architecture of the Laravel app"
+layout="responsive"
+width={3118} height={1776} quality={100} />
+_My Majestic Monolith Laravel app_
+
+Please follow these steps:
```
coffee-cup, line 73: I think the phrasing here could be improved a bit. As a user reading this I'm a bit unsure _why_ I need to make these scripts. I think the guide could do a better job explaining what is actually happening. Right now it feels a bit like "trust me just do this".

src/docs/guides/laravel.md
```diff
@@ -0,0 +1,185 @@
...
+            # Make sure this file has executable permissions, run `chmod +x run-worker.sh`
+
+            # Run the queue worker
+            php artisan queue:work     
+        ```
+    -  Add the content below to the `run-cron.sh` file:
+        ```bash
+            #!/usr/bin/env bash
+
+            # Make sure this file has executable permissions, run `chmod +x run-cron.sh`
```
coffee-cup, line 112: Can we use Railway crons for this?

src/docs/guides/laravel.md
unicodeveloper: Good point!

unicodeveloper:
> The formatting here is a bit strange image

Fixed

src/docs/guides/laravel.md
unicodeveloper: Fixed this

src/docs/guides/laravel.md
```diff
@@ -0,0 +1,187 @@
...
+We highly recommend that [you eject from the template after deployment](/guides/deploy#eject-from-template-repository) to create a copy of the repo on your GitHub account.
+
+**Note:** If you see a "500 Server Error" on your deployed URL, it’s likely because Laravel is missing the **APP_KEY** or other required environment variables. To fix this, go to the Railway dashboard and add the necessary environment variables—especially the **APP_KEY**—and then redeploy the application. You can generate a new **APP_KEY** using this [Laravel APP KEY generator](https://generate-random.org/laravel-key-generator).
+
+
+## Deploy from a GitHub Repo
+
+To deploy a Laravel app on Railway directly from GitHub, follow the steps below:
+
+1. Fork the basic <a href="https://github.com/railwayapp-templates/laravel" target="_blank">Laravel GitHub repo</a>.
```
coffee-cup, line 29: What are your thoughts on having the guide start from a fresh Laravel app? Instead of forking an existing repo, it could start from new `composer create-project laravel/laravel my-project`?

What does the template have that is different from starting from scratch?

unicodeveloper: It’s a combination of a few factors:

**Different Methods for Setting Up a Laravel App**:

- Users might be using tools like [Laravel Herd](https://herd.laravel.com/), which provides a UI for creating new projects. For instance, I use Herd a lot. 
- Alternatively, they could be using the [Laravel Installer](https://github.com/laravel/installer), where running laravel new <sample-app> in the terminal creates a new app.
For composer setups, I’ll need to add instructions to ensure the reader has composer installed.

**Starting from Scratch Can Be Overly Verbose**:

Building a fresh Laravel app from scratch makes the guide a lot more verbose which defeats the goal of reducing the time to deploy. For example, if using the "Deploy from GitHub" option (which I know is the most popular choice), I’d first need to guide the user through pushing their code to GitHub before proceeding with deployment instructions.

Instead, using a fork provides a quick and easy way to get a copy of the app in less than 30 seconds, saving time and making the user see it quickly on the Railway new project modal. This is why I also included an option to skip these steps if the user already has a GitHub repository.

brody192: I have to agree with jr here, most people looking for a place to host their Laravel app would already have a project locally or on GitHub and wouldn't be interested in starting from scratch, so starting from the most common way to provision a Laravel app would give them a better understanding of what they need to do to their pre-existing local project so that deployment to Railway is a smooth process.

coffee-cup: I think the intended audience of this guide is someone who already knows Laravel and has a project that they want to host on Railway. We should assume some Laravel knowledge and explain how to specifically host on Railway with best practices.

unicodeveloper: Like we all agree. This guide is for people that already know Laravel or have used it at least once and just want to host on Railway so it means they already have a running app locally or on GitHub. 

I just want to reiterate the following about the step by step approach in the guide:

- The GitHub repo option shows you how to deploy your Laravel app if it's on GitHub.
- The Deploy from CLI option shows you how to deploy your Laravel app if you have it locally like most people do. (if you don't want to push to GitHub).
- The Deploy via Custom scripts option uses both approach. It starts from locally and moves to having it on GitHub.

I'll just remove the fork option. I think that's the part that's making this debatable.

unicodeveloper:
Please let me know what you think about the new changes I made @coffee-cup @brody192

src/docs/guides/laravel.md
unicodeveloper: I also changed the last section to **Set Up Database, Migrations, Crons and Workers** to logically take the reader from the initial steps to setting up all the other services for their Laravel apps to functional optimally.

src/docs/guides/laravel.md
```diff
@@ -0,0 +1,199 @@
...
+
+## One-Click Deploy from a Template
+
+[![Deploy on Railway](https://railway.app/button.svg)](https://railway.app/new/template/fWEWWf)
+
+This template sets up a basic Laravel application along with a Postgres database on Railway. You can also choose from a <a href="https://railway.app/templates?q=laravel" target="_blank">variety of Laravel app templates</a> created by the community.
+
+We highly recommend that [you eject from the template after deployment](/guides/deploy#eject-from-template-repository) to create a copy of the repo on your GitHub account.
+
+**Note:** If you see a "500 Server Error" on your deployed URL, it’s likely because Laravel is missing the **APP_KEY** or other required environment variables. To fix this, go to the Railway dashboard and add the necessary environment variables—especially the **APP_KEY**—and then redeploy the application. You can generate a new **APP_KEY** using this [Laravel APP KEY generator](https://generate-random.org/laravel-key-generator).
```
brody192, line 21: Is there any reason the template doesn't set this up for the user? 

https://docs.railway.app/guides/create#template-variable-functions

src/docs/guides/laravel.md
```diff
@@ -0,0 +1,199 @@
...
+_My Majestic Monolith Laravel app_
+
+
+Please follow these steps to get started:
+
+1. Create three bash scripts in the root directory of your Laravel app directory: `run-app.sh`, `run-worker.sh`, and `run-cron.sh`. 
+
+    These scripts will contain the commands needed to deploy and run the app, worker, and cron services for your Laravel app on Railway. 
+    
+    After creating the files, make them executable by running `chmod +x` on each file in the terminal. This will allow your scripts to run properly when executed.
```
brody192, line 92: Nit: What if I'm on Windows?

In the build or start commands, Instead of `sh /script.sh` could we do `chmod +x ./script && sh ./script.sh`

src/docs/guides/laravel.md
unicodeveloper: It doesn't. This is why I was asking you the other day if I could modify the template from the template composer. @brody192

src/docs/guides/laravel.md
unicodeveloper: Oh, that's smart. So instead of doing it locally, the server does that for us when it's about to invoke it? 

But wouldn't that be repetitive? because it means every time we deploy, it always has to run the change permission command. that seems weird 🤔

src/docs/guides/laravel.md
brody192: Okay, I'll try to get you in touch with the creator.

src/docs/guides/laravel.md
brody192: Yeah, but I would rather that than answer a support thread on why the script doesn't have execution permissions.

src/docs/guides/laravel.md
```diff
@@ -0,0 +1,199 @@
...
+2. Click **Deploy from GitHub repo**.
+3. Select your GitHub repo.
+    - Railway requires a valid GitHub account to be linked. If your Railway account isn't associated with one, you will be prompted to link it.
+4. Click **Add Variables**. 
+    - Add all your app environment variables.
+5. Click **Deploy**.
+
+Once the deployment is successful, a Railway [service](/guides/services) will be created for you. By default, this service will not be publicly accessible.
+
+**Note:** Railway will automatically detect that it's a Laravel app during [deploy and run your app via php-fpm and nginx](https://github.com/railwayapp/nixpacks/tree/main/src/providers/php).
```
aleksrutins, line 37: It might be better to link to [the Nixpacks docs](https://nixpacks.com/docs/providers/php) here instead of linking to the source code.

aleksrutins (changes requested):

src/docs/guides/laravel.md
unicodeveloper: Thanks for reaching out @brody192.

src/docs/guides/laravel.md
unicodeveloper: Yeah that makes sense. Thanks for the suggestion @aleksrutins

aleksrutins (approved):

PR #574, opened 2024-10-04
Languages & Frameworks: Add Ruby on Rails guide (Ruby)

unicodeveloper:
Completes Linear Issue [#LOG-2577](https://linear.app/railway/issue/LOG-2577).

CleanShot 2024-10-04 at 16 00 18@2x

### DIFF
--- a/src/docs/guides/rails.md
+++ b/src/docs/guides/rails.md
@@ -1,3 +1,153 @@
 ---
 title: Deploy a Ruby on Rails App
 ---
+
+Rails is a Ruby full-stack framework designed to compress the complexity of modern web apps. It ships with all the tools needed to build amazing web apps on both the front and back end.
+
+## Create a Rails App
+
+**Note:** If you already have a Rails app locally or on GitHub, you can skip this step and go straight to the [Deploy Ruby on Rails App on Railway](#deploy-ruby-on-rails-app-on-railway).
+
+To create a new Rails app, ensure that you have [Ruby](https://www.ruby-lang.org/en/documentation/installation) and [Rails](https://guides.rubyonrails.org/getting_started.html#creating-a-new-rails-project-installing-rails-installing-rails) installed on your machine. Once everything is set up, run the following command in your terminal:
+
+```bash
+rails new blog --database=postgresql
+```
+
+This command will create a new Rails app named `blog` with PostgreSQL as the database config. Now, let’s create a simple "Hello World" page to ensure everything is working correctly.
+
+1. **Generate a Controller**: Run the following command to create a new controller named `HelloWorld` with an `index` action:
+    ```bash
+        rails g controller HelloWorld index
+    ```
+
+    This will generate the necessary files for the controller, along with a view, route, and test files.
+
+2. **Update the Routes File**: Open the `config/routes.rb` file and modify it to set the root route to the `hello_world#index` action:
+
+    ```ruby
+    Rails.application.routes.draw do
+        get "hello_world/index"
+        # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
+
+        # Reveal health status on /up that returns 200 if the app boots with no exceptions, otherwise 500.
+        # Can be used by load balancers and uptime monitors to verify that the app is live.
+        get "up" => "rails/health#show", as: :rails_health_check
+
+        # Render dynamic PWA files from app/views/pwa/*
+        get "service-worker" => "rails/pwa#service_worker", as: :pwa_service_worker
+        get "manifest" => "rails/pwa#manifest", as: :pwa_manifest
+
+        # Defines the root path route ("/")
+        root "hello_world#index"
+    end
+    ```
+3. **Modify the View**: Open the `app/views/hello_world/index.html.erb` file and replace its content with the following:
+
+    ```ruby
+    <h1>Hello World</h1>
+
+    <p> This is a Rails app running on Railway</p>
+    ```
+4. **Run the Application Locally**: Start the Rails server by running:
+
+    ```bash
+        bin/rails server
+    ```
+
+    Open your browser and go to `http://localhost:3000` to see your "Hello World" page in action.
+
+Now that your app is running locally, let’s move on to deploying it to Railway!
+
+## Deploy Ruby on Rails App on Railway
+
+Railway offers multiple ways to deploy your Rails app, depending on your setup and preference. Choose any of the following methods:
+
+1. [One-click deploy from a template](#one-click-deploy-from-a-template).
+2. [Using the CLI](#deploy-from-the-cli).
+3. [From a GitHub repository](#deploy-from-a-github-repo).
+
+## One-Click Deploy from a Template
+
+If you’re looking for the fastest way to get started, the one-click deploy option is ideal. It sets up a Rails app along with a Postgres database and Redis.
+
+Click the button below to begin:
+
+[![Deploy on Railway](https://railway.app/button.svg)](https://railway.app/new/template/sibk1f)
+
+After deploying, we recommend that you [eject from the template](/guides/deploy#eject-from-template-repository) to create a copy of the repository under your own GitHub account. This will give you full control over the source code and project.
+
+## Deploy from the CLI
+
+To deploy the Rails app using the Railway CLI, please follow the steps:
+
+1. **Install the Railway CLI**:
+    - <a href="/guides/cli#installing-the-cli" target="_blank">Install the CLI</a> and <a href="/guides/cli#authenticating-with-the-cli" target="_blank">authenticate it</a> using your Railway account.
+2. **Initialize a Railway Project**:
+    - Run the command below in your Rails app directory. 
+        ```bash
+        railway init
+        ```
+    - Follow the prompts to name your project.
+    - After the project is created, click the provided link to view it in your browser.
+3. **Deploy the Application**:
+    - Use the command below to deploy your app:
+        ```bash
+        railway up
+        ```
+    - This command will scan, compress and upload your app's files to Railway. You’ll see real-time deployment logs in your terminal.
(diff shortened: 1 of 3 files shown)


unicodeveloper:
I added the section to set up workers and crons with Sidekiq. So the guide is very complete now.

@brody192 please can you review. Would love to get this merged in before the weekend.

src/docs/guides/rails.md
```diff
@@ -0,0 +1,200 @@
...
+1. **Generate a Controller**: Run the following command to create a new controller named `HelloWorld` with an `index` action:
+    ```bash
+        rails g controller HelloWorld index
+    ```
+
+    This will generate the necessary files for the controller, along with a view, route, and test files.
+
+2. **Update the Routes File**: Open the `config/routes.rb` file and modify it to set the root route to the `hello_world#index` action:
+
+    ```ruby
```
brody192, line 28: Maybe you could add syntax highlighting for ruby?

src/docs/guides/rails.md
```diff
@@ -0,0 +1,200 @@
...
+7. **Verify the Deployment**:
+    - Once the deployment completes, go to **View logs** to check if the server is running successfully.
+
+**Note:** If your app has a `Dockerfile` (which newer Rails apps typically include by default), Railway will [automatically detect and use it to build](/reference/dockerfiles) your app. If not, Railway will still handle the deployment process for you.
+
+8. **Set Up a Public URL**:
+    - Navigate to the **Networking** section under the [Settings](/overview/the-basics#service-settings) tab of your new service.
+    - Click [Generate Domain](/guides/public-networking#railway-provided-domain) to create a public URL for your app.
+
+## Deploy from a GitHub Repo
```
brody192, line 120: Curious, why does the CLI guide mention setting up Postgres but the Github guide doesn't?

src/docs/guides/rails.md
```diff
@@ -0,0 +1,200 @@
...
+        ```
+    - Follow the prompts to name your project.
+    - After the project is created, click the provided link to view it in your browser.
+3. **Deploy the Application**:
+    - Use the command below to deploy your app:
+        ```bash
+        railway up
+        ```
+    - This command will scan, compress and upload your app's files to Railway. You’ll see real-time deployment logs in your terminal.
+ - **Note:** If you see an error about a missing `secret_key_base` for the production environment, don’t worry. We’ll fix this in the next step.
```
brody192, line 99: Instead of having the user deploy their app before the database, have them deploy their database before the app so that they can setup the needed variables on the service before initial deploy, this would prevent the unnecessary crash and redeploy.

src/docs/guides/rails.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. **Install Sidekiq**
+    - Start by adding `sidekiq` and `sidekiq-cron` to your Rails app. In your terminal, run the following command:
+        ```bash
+        bundle add sidekiq
+        bundle add sidekiq-cron
+        ```
+2. **Add a Redis Database Service**
+    - Sidekiq uses Redis as a job queue. To set this up:
+        - Right-click on the Railway dashboard canvas or click the **Create** button.
```
brody192, line 160: project canvas*

src/docs/guides/rails.md
```diff
@@ -0,0 +1,200 @@
...
+        - Select **Database**.
+        - Select **Add Redis** from the available databases.
+            - This will create and deploy a new Redis service for your app.
+3. **Create and Configure a Worker Service**
+    - Now, set up a separate service to run your Sidekiq workers.
+        - Create a new **Empty Service** and name it **Worker Service**.
```
brody192, line 166: How?

src/docs/guides/rails.md
```diff
@@ -0,0 +1,200 @@
...
+    - This command will scan, compress and upload your app's files to Railway. You’ll see real-time deployment logs in your terminal.
+ - **Note:** If you see an error about a missing `secret_key_base` for the production environment, don’t worry. We’ll fix this in the next step.
+4. **Add a Database Service**:
+    - Run `railway add`.
+    - Select `PostgreSQL` by pressing space and hit **Enter** to add it to your project.
+    - A database service will be added to your Railway project.
+5. **Configure Environment Variables**:
+    - Go to your app service <a href="/overview/the-basics#service-variables">**Variables**</a> section and add the following:
+        - `SECRET_KEY_BASE` or `RAILS_MASTER_KEY`: Set the value to the key from your local app's `config/master.key`.
+        - `DATABASE_URL`: Set the value to `${{Postgres.DATABASE_PUBLIC_URL}}` (this references the URL of your new Postgres database). Learn more about [referencing service variables](/guides/variables#referencing-another-services-variable).
```
brody192, line 107: Why the public URL?

src/docs/guides/rails.md
```diff
@@ -0,0 +1,200 @@
...
+        - Go to the <a href="/overview/the-basics#service-settings">**Settings**</a> tab of this service to configure it.
+        - In the **Source** section, connect your GitHub repository to the **Source Repo**.
+        - Under the <a href="/guides/build-configuration#customize-the-build-command">**Build**</a> section, set `bundle install` as the **Custom Build Command**. This installs the necessary dependencies.
+        - In the **Deploy** section, set `bundle exec sidekiq` as the <a href="/guides/start-command">**Custom Start Command**</a>. This command will start Sidekiq and begin processing jobs.
+        - Click on <a href="/overview/the-basics#service-variables">**Variables**</a> at the top of the service settings.
+        - Add the following environment variables:
+            - `SECRET_KEY_BASE` or `RAILS_MASTER_KEY`: Set this to the value of your Rails app’s secret key.
+            - `REDIS_URL`: Set this to `${{Redis.REDIS_PUBLIC_URL}}` to reference the Redis database URL. This tells Sidekiq where to find the job queue. Learn more about [referencing service variables](/guides/variables#referencing-another-services-variable).
```
brody192, line 174: Public URL?

src/docs/guides/rails.md
unicodeveloper: That would be nice!

src/docs/guides/rails.md
unicodeveloper: Remember the convo we had about talking to the database using private networking? It doesn't work during database migration on every deploy so I simply use the public URL which works seamlessly.

src/docs/guides/rails.md
unicodeveloper: Mentioned the reason below

src/docs/guides/rails.md
unicodeveloper: It doesn't matter though. We are deploying via the CLI, even if the database is set up first via the cli, the app service would not exist for the reader to be able to add variables to it before `railway up`.

`railway up` is what creates the service. The only way that it will work is if the readers goes to the project canvas to create an empty service first and then  add the variables. And that entirely defeats the purpose of using the CLI for the first time.

src/docs/guides/rails.md
unicodeveloper: I omitted it because that step is already mentioned in the CLI guide. So, it assumes the reader already know to do that from reading the CLI guide (which is the first step they encounter in the guide). Do you think adding it here also matters? 🤔

src/docs/guides/rails.md
unicodeveloper: I'll link to it

src/docs/guides/rails.md
brody192: Yeah, since If someone is not interested in the CLI guide they might outright skip it.

src/docs/guides/rails.md
brody192: Fair point, maybe the CLI should support creating empty services, but that's out of scope.

src/docs/guides/rails.md
brody192: Can Rails be made to do migrations during runtime before starting the app?

src/docs/guides/rails.md
unicodeveloper: Yes out of scope. 

One thing we can do later is to improve the CLI is add a feature/command to be able to add/set/edit environment variables from the CLI. We currently don't have that option. I used this heavily in [Heroku's CLI ](https://devcenter.heroku.com/articles/config-vars#managing-config-vars)a couple of years back.

src/docs/guides/rails.md
unicodeveloper: I'll try this out and revert

unicodeveloper: This works perfectly on runtime! Updated the guide to have the relevant knowledge. @brody192

src/docs/guides/rails.md
unicodeveloper: Added.

src/docs/guides/rails.md
unicodeveloper: Added for this and a couple of languages

unicodeveloper:
@brody192 Please one more check so I can merge it in.

brody192:
Maybe you could add a section on how you could open a REPL, otherwise LGTM!

unicodeveloper:
> Maybe you could add a section on how you could open a REPL, otherwise LGTM!

Open a REPL? I know what a REPL mean, but you clarify what you mean in relation to the guide? @brody192

brody192:
I've seen a decent amount of people want to know how to do that for a rails app, will it never be needed when developing a rails app on Railway?

unicodeveloper:
> I've seen a decent amount of people want to know how to do that for a rails app, will it never be needed when developing a rails app on Railway?

No it won't. Mostly needed if you are still building your app, and you are trying out a lot of things like helper functions, active record interactions etc.

brody192:
Gotcha, thanks for explaining.

PR #579, opened 2024-10-16
Add Django guide

unicodeveloper:
CleanShot 2024-10-17 at 00 54 38@2x

Might add a section on running Celery tomorrow!

### DIFF
--- a/src/docs/guides/django.md
+++ b/src/docs/guides/django.md
@@ -1,0 +1,254 @@
+---
+title: Deploy a Django App
+---
+
+[Django](https://www.djangoproject.com) is a powerful Python web framework that simplifies web development by providing ready-to-use tools for rapid development and clean design. 
+
+It’s free, open-source, and comes with a range of features to streamline tasks like authentication, routing, and database management, so developers can focus on building their applications without handling everything from scratch.
+
+## Create a Django App
+
+**Note:** If you already have a Django app locally or on GitHub, you can skip this step and go straight to the [Deploy Django App on Railway](#deploy-django-app-on-railway).
+
+To create a new Django app, ensure that you have [Python](https://www.python.org/downloads/) and [Django](https://docs.djangoproject.com/en/5.1/intro/install/) installed on your machine.
+
+Follow the steps blow to set up the project in a directory.
+
+Create a virtual environment
+```bash
+python -m venv env
+```
+
+Activate the virtual environment
+```bash
+source env/bin/activate
+```
+
+Install Django
+```bash
+python -m pip install django
+```
+
+Once everything is set up, run the following command in your terminal to provision a new Django project:
+
+```bash
+django-admin startproject liftoff
+```
+
+This command will create a new Django project named `liftoff`. 
+
+Next, `cd` into the directory and run `python manage.py runserver` to start your project.
+
+Open your browser and go to `http://127.0.0.1:8000` to see the project. You'll see the Django welcome page with a "The install worked successfully! Congratulations!" paragraph.
+
+**Note:** You'll see a red notice about unapplied migration(s). You can ignore them for now. We'll run them when we deploy the project.
+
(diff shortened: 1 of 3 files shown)


src/docs/guides/django.md
```diff
@@ -0,0 +1,254 @@
...
+Follow the steps blow to set up the project in a directory.
+
+Create a virtual environment
+```bash
+python -m venv env
+```
+
+Activate the virtual environment
+```bash
+source env/bin/activate
```
brody192, line 24: What about windows developers?

src/docs/guides/django.md
```diff
@@ -0,0 +1,254 @@
...
+6. Create a **static** folder:
+
+Inside your `liftoff` directory, create a static folder where all static assets will reside.
+
+7. Create a `requirements.txt` file: 
+ 
+To track all the dependencies for deployment, create a `requirements.txt` file:
+
+```bash
+pip freeze > requirements.txt
```
brody192, line 140: maybe mention that this is only safe to run since we are in a venv, otherwise this will freeze all packages the user has ever installed on their system

src/docs/guides/django.md
```diff
@@ -0,0 +1,254 @@
...
+    - Select **Deploy from GitHub repo** and choose your repository.
+        - If your Railway account isn’t linked to GitHub yet, you’ll be prompted to do so.
+3. **Add Environment Variables**:
+    - Click **Add Variables** and configure all the necessary environment variables for your app.
+        - `PGDATABASE`: Set the value to `${{Postgres.PGDATABASE}}` (this references the Postgres database name). Learn more about [referencing service variables](/guides/variables#referencing-another-services-variable).  
+        - `PGUSER`: Set the value to `${{Postgres.PGUSER}}`
+        - `PGPASSWORD`: Set the value to `${{Postgres.PGPASSWORD}}`
+        - `PGHOST`: Set the value to `${{Postgres.PGHOST}}`
+        - `PGPORT`: Set the value to `${{Postgres.PGPORT}}`
+    **Note:** We don't have the Postgres Database service yet. We'll add that soon.
```
brody192, line 226: when rendered this line appears on the line above

src/docs/guides/django.md
```diff
@@ -0,0 +1,254 @@
...
+4. **Deploy the App**: 
+    - Click **Deploy** to start the deployment process.
+    - Once the deployed, a Railway [service](/guides/services) will be created for your app, but it won’t be publicly accessible by default.
+    **Note:** You'll encounter an error about the PGDATABASE environment not set. Don't worry, we'll fix that soon.
+5. **Add a Database Service**:
```
brody192, line 231: i know the add environment variables mode is designed to add variables, but is there anything stopping you from also adding a database in that time too?

src/docs/guides/django.md
unicodeveloper: Good point!

src/docs/guides/django.md
unicodeveloper: Oh this is why I started the venv because I knew we will get to this point. I'll add that as a note.

src/docs/guides/django.md
unicodeveloper: That's weird!

src/docs/guides/django.md
unicodeveloper: Actually that's a good idea, because the database will still get provisioned if I add it.  I'll move the step to add the database up

unicodeveloper: Updated!

src/docs/guides/django.md
unicodeveloper: Added the note.

src/docs/guides/django.md
unicodeveloper: Added a note.

PR #588, opened 2024-10-25
Add Rust Rocket Guide

unicodeveloper:
Rocket Guide

CleanShot 2024-10-25 at 15 43 41@2x

### DIFF
--- a/src/docs/guides/rocket.md
+++ b/src/docs/guides/rocket.md
@@ -1,0 +1,188 @@
+---
+title: Deploy a Rust Rocket App
+---
+
+[Rocket](https://rocket.rs) is a web framework for Rust that makes it simple to write fast, type-safe, secure web applications with incredible usability, productivity and performance.
+
+This guide covers how to deploy a Rocket app to Railway in four ways:
+
+1. [One-click deploy from a template](#one-click-deploy-from-a-template).
+2. [From a GitHub repository](#deploy-from-a-github-repo).
+3. [Using the CLI](#deploy-from-the-cli).
+4. [Using a Dockerfile](#use-a-dockerfile).
+
+Now, let's create a Rocket app! 🚀
+
+## Create a Rocket App
+
+**Note:** If you already have a Rocket app locally or on GitHub, you can skip this step and go straight to the [Deploy Rocket App to Railway](#deploy-the-rocket-app-to-railway).
+
+To create a new Rocket app, ensure that you have [Rust](https://www.rust-lang.org/tools/install) installed on your machine.
+
+Run the following command in your terminal to create a new Rust app:
+
+```bash
+cargo new helloworld --bin
+```
+
+The command creates a new binary-based Cargo project in a `helloworld` directory.
+
+Next, `cd` into the directory and add Rocket as a dependency in your `Cargo.toml`:
+
+```bash
+[dependencies]
+rocket = "0.5.1"
+```
+
+### Modify the Application File
+
+Next, open the app in your IDE and navigate to the `src/main.rs` file.
+
+Replace the content with the code below:
+
+```rust
+#[macro_use] 
+extern crate rocket;
+
+#[get("/")]
+fn index() -> &'static str {
+    "Hello world, Rocket!"
+}
+
+#[launch]
+fn rocket() -> _ {
+    rocket::build().mount("/", routes![index])
+}
+```
+
+The code above uses the Rocket framework to create a basic web server that responds to HTTP requests. It defines a simple route using the `#[get("/")]` macro, which tells Rocket to handle GET requests to the root URL `(/)`. 
+
+The `index()` function is the handler for this route and returns a static string, **"Hello world, Rocket!"**, which will be sent as the response when the root URL is accessed.
+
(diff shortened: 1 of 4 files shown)


src/docs/guides/rocket.md
```diff
@@ -0,0 +1,188 @@
...
+
+Run the following command in your terminal to create a new Rust app:
+
+```bash
+cargo new helloworld --bin
+```
+
+The command creates a new binary-based Cargo project in a `helloworld` directory.
+
+Next, `cd` into the directory and add Rocket as a dependency in your `Cargo.toml`:
```
brody192, line 30: Isn't there some cargo command that adds dependencies to the Cargo.toml file for you? similar to `go get`

src/docs/guides/rocket.md
```diff
@@ -0,0 +1,188 @@
...
+
+To deploy a Rocket app to Railway directly from GitHub, follow the steps below:
+
+1. **Create a New Project on Railway**:
+    - Go to <a href="https://railway.app/new" target="_blank">Railway</a> to create a new project.
+2. **Deploy from GitHub**: 
+    - Select **Deploy from GitHub repo** and choose your repository.
+        - If your Railway account isn’t linked to GitHub yet, you’ll be prompted to do so.
+3. **Add Environment Variables**:
+    - Click **Add Variables**, then add **ROCKET_ADDRESS** with the value `0.0.0.0`. This allows your Rocket app to accept external connections by listening on `0.0.0.0`.
```
brody192, line 142: IMO `ROCKET_ADDRESS` should be in a code block.

src/docs/guides/rocket.md
```diff
@@ -0,0 +1,188 @@
...
+    - Once the deployment completes, go to **View logs** to check if the server is running successfully.
+
+    **Note:** During the deployment process, Railway will automatically [detect that it’s a Rust app](https://nixpacks.com/docs/providers/rust).
+5. **Set Up a Public URL**:
+    - Navigate to the **Networking** section under the [Settings](/overview/the-basics#service-settings) tab of your new service.
+    - Click [Generate Domain](/guides/public-networking#railway-provided-domain) to create a public URL for your app.
+
+### Use a Dockerfile
+
+1. Create a `Dockerfile` in the `helloworld` or Rocket app's root directory.
```
brody192, line 156: I've always been told cargo chef is so much better, maybe we should show what a cargo chef Dockerfile looks like?

src/docs/guides/rocket.md
unicodeveloper: True

src/docs/guides/rocket.md
unicodeveloper: Interesting. Let me check that instead.

src/docs/guides/rocket.md
unicodeveloper: Yes there is, but you need to install the cargo-edit package first. Then you can now use `cargo add`. 

For this guide, it feels like an overkill since we are adding just `rocket` as the dependency.

brody192: Oh wow, needing to install an extra package to be able to add packages, Rust is so amazing!

unicodeveloper: Lmao

unicodeveloper:
@brody192 Added the Axum guide and also did an update to use cargo-chef

src/docs/guides/axum.md
```diff
@@ -0,0 +1,191 @@
...
+    Router,
+};
+
+#[tokio::main]
+async fn main() {
+    // build our application with a single route
+    let app = Router::new().route("/", get(root));
+
+    // run our app with hyper, listening globally on port 3000
+    let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();
```
brody192, line 58: The application should listen on `$PORT`

unicodeveloper: It's already listening on a default PORT of 3000. So is there something I am missing? 🤔

unicodeveloper: This part of the guide is for the reader to run it locally which means they can use any port. 

For deployment, it ignores it and uses the default port on the V2 runtime. So does it still make sense to emphasize changing it here when it really doesn't matter?

brody192: I don't know rust, but how does it listen on $PORT when on Railway?

But yes, you would want to do something like what you do in node -

`process.env.PORT || 3000`

Here is how the template does it -

https://github.com/railwayapp-templates/rust-starter/blob/main/src%2Fmain.rs#L36

unicodeveloper: Oh I get you, yeah that makes sense. I will make some changes.

src/docs/guides/rocket.md
Milo123459: No, you don't. In new rust versions cargo add is a regular command.

unicodeveloper: Thanks, just discovered that worked. Looks I was using a really older version

src/docs/guides/axum.md
unicodeveloper: Fixed.

PR #597, opened 2024-11-01
Add Flask guide

unicodeveloper:
CleanShot 2024-11-01 at 16 27 49@2x

### DIFF
--- a/src/docs/guides/flask.md
+++ b/src/docs/guides/flask.md
@@ -1,0 +1,221 @@
+---
+title: Deploy a Flask App
+---
+
+[Flask](https://flask.palletsprojects.com/en/stable) is a Python micro framework for building web applications.
+
+This guide covers how to deploy a Flask app to Railway in four ways:
+
+1. [One-click deploy from a template](#one-click-deploy-from-a-template).
+2. [From a GitHub repository](#deploy-from-a-github-repo).
+3. [Using the CLI](#deploy-from-the-cli).
+4. [Using a Dockerfile](#use-a-dockerfile).
+
+Now, let's create a Flask app!
+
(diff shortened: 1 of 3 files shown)


brody192:
Looks like you didn't update sidebar.ts

unicodeveloper:
> Looks like you didn't update sidebar.ts

Oops missed that.

src/docs/guides/flask.md
```diff
@@ -0,0 +1,221 @@
...
+    - Click [Generate Domain](/guides/public-networking#railway-provided-domain) to create a public URL for your app.
+
+## Use a Dockerfile
+
+1. Create a `Dockerfile` in the app's root directory.
+2. Add the content below to the `Dockerfile`:
+    ```docker
+    # Use the Python 3 alpine official image
+    # https://hub.docker.com/_/python
+    FROM python:3-alpine
```
brody192, line 194: We try to shy away from alpine images, they have a strange networking stack and have shown to have lower CPU performance compared to the Debian counterparts.

src/docs/guides/flask.md
```diff
@@ -0,0 +1,221 @@
...
+
```
brody192, line 195: We must add `ENV PYTHONUNBUFFERED=1` here, if this is omitted log lines that get buffered could be dropped, this is something nixpacks does by default for python.

unicodeveloper: Oh, that's new knowledge.

src/docs/guides/flask.md
unicodeveloper: Interesting...never knew that.

PR #606, opened 2024-11-13
Add Deploy Express Guide

unicodeveloper:
CleanShot 2024-11-13 at 15 20 50@2x

### DIFF
--- a/src/docs/guides/express.md
+++ b/src/docs/guides/express.md
@@ -1,0 +1,223 @@
+---
+title: Deploy an Express App
+---
+
+[Express](https://expressjs.com) is a fast and flexible web application framework for Node.js that provides a simple and minimalistic approach to building web servers and APIs. It is known for its speed and unopinionated nature, allowing developers to structure their applications as they see fit while offering powerful features.
+
+This guide covers how to deploy an Express app to Railway in four ways:
+
+1. [One-click deploy from a template](#one-click-deploy-from-a-template).
+2. [From a GitHub repository](#deploy-from-a-github-repo).
+3. [Using the CLI](#deploy-from-the-cli).
+4. [Using a Dockerfile](#use-a-dockerfile).
+
+Now, let's create an Express app!
+
+## Create an Express App
+
+**Note:** If you already have an Express app locally or on GitHub, you can skip this step and go straight to the [Deploy Express App to Railway](#deploy-express-app-to-railway).
+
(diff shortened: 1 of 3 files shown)


src/docs/guides/express.md
```diff
@@ -0,0 +1,229 @@
...
+**Note:** If you already have an Express app locally or on GitHub, you can skip this step and go straight to the [Deploy Express App to Railway](#deploy-express-app-to-railway).
+
+To create a new Express app, ensure that you have [Node](https://nodejs.org/en/learn/getting-started/how-to-install-nodejs) installed on your machine.
+
+Create a directory, `helloworld`, and `cd` into it.
+
+Run the following command in your terminal to create a new Express app within the `helloworld` directory:
+
+```bash
+npx express-generator --view=pug
```
brody192, line 27: Why would a user want to use pug?

why would that be something we recommend to them?

Genuine question, since for most people I imagine they use express for backend, and react for frontend.

src/docs/guides/express.md
```diff
@@ -0,0 +1,229 @@
...
+1. [One-click deploy from a template](#one-click-deploy-from-a-template).
+2. [From a GitHub repository](#deploy-from-a-github-repo).
+3. [Using the CLI](#deploy-from-the-cli).
+4. [Using a Dockerfile](#use-a-dockerfile).
+
+Now, let's create an Express app!
+
+## Create an Express App
+
+**Note:** If you already have an Express app locally or on GitHub, you can skip this step and go straight to the [Deploy Express App to Railway](#deploy-express-app-to-railway).
```
brody192, line 18: Hash should be `#deploy-the-express-app-to-railway`

src/docs/guides/express.md
unicodeveloper: Good question.

I know a lot of people use express for backend, but I wanted the guide to show something more fullstack-ish. A different approach. 

The default express generator uses Jade, but then it recommends that people use Pug. That's why I am recommending Pug.

unicodeveloper: Maybe I can just make it API alone then? Thoughts?

brody192: Fair enough!

unicodeveloper: Fair enough to do the template recommendation thing or to turn it into API alone? 🤔

brody192: I don't think we have an express Pug template, so yeah why not make a template out of it!

unicodeveloper: Fixed!

PR #643, opened 2024-12-21
Document how to authenticate with public API using a project key

pascal-giguere:
Current docs mention project keys as one of the key types supported by the public API, but doesn't mention how to use them.

This clarifies that requests authenticated with a project key must use a `Project-Access-Token` header (rather than `Authorization`), as found in the CLI's code [here](https://github.com/railwayapp/cli/blob/master/src/client.rs#L20).


brody192:
Thank you for the PR, this definitely makes it more clear!

Just had some minor fixed, hope you didn't mind.

PR #662, opened 2025-01-23
Fix broken links

FraglyG:
- CLI installation references were didn't have a `/` prefix
- Mongo Replica Set template's URL changed to ha-mongo

### DIFF
--- a/src/docs/guides/build-a-database-service.md
+++ b/src/docs/guides/build-a-database-service.md
@@ -77,5 +77,5 @@ Here are some suggestions to check out -
 - <a href="https://railway.com/template/SMKOEA" target="_blank">Minio</a>
 - <a href="https://railway.com/template/clickhouse" target="_blank">ClickHouse</a>
 - <a href="https://railway.com/template/dragonfly" target="_blank">Dragonfly</a>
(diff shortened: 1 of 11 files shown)


brody192:
I wonder if the other HA template links are outdated given I did change them all to `ha-*` after they where made public.

FraglyG:
I just checked and couldn't find any other broken HA template links, not sure if you fixed them since you've said that
I did find some other broken links though, I'll add them to this PR later

FraglyG:
I took the liberty of going through all of the template links I could find (I might've missed some but I covered most of them) to make sure they're all in tact, it seems they all are! 

I did notice however that some pages reference the outdated Chroma template, would it be fine if I replaced those links with my Chroma template? It's more up-to-date and maintained.

Pages referencing outdated Chroma template:
- https://docs.railway.com/reference/databases
- https://docs.railway.com/guides/build-a-database-service

brody192:
Yep you are free to replace the template with your template, and update the https://railway.app links!

FraglyG:
Awesome! Thank you :D
I believe that should be all then

brody192:
Thank you!

PR #684, opened 2025-02-27
Clarify private networking PORT reference

neallseth:
Clarifying reference variable usage in private networking


brody192:
Just needed to update it to use the React component for the banner instead of the markdown version, but it's worded perfectly!

Thank you for the contribution and sorry the review took so long!

PR #703, opened 2025-03-12
Trial Accounts: Make explicit the 5 services limitation

echohack:
While I was working during the Free Trial, I ran into a limitation where any project that I made could only have a maximum of 5 services (even empty services). I would receive an error "Failed to deploy" in the toast notification, but no other information was available.

I upgraded to the Hobby plan and this error immediately went away.

There should probably be a mention of this limitation in the relevant documentation, and it might be helpful to explicitly have an error message for Trial accounts should they reach this limitation.

If this limitation is instead a bug that's not intended, then please feel free to decline the PR (and likely open a bug internally for whatever exactly the issue might be)


ndneighbor:
Yup.... great catch! Merging

PR #718, opened 2025-05-02
Update next guide location?

ColtonIdle:
Next was the only one that went to quick-start. Not sure if that was intentional?


echohack:
Intentional. Thanks for noticing the inconsistency in routing though!

PR #790, opened 2025-07-01
Update free-trial.md

anre2025:


brody192:
Limited Trial users cannot deploy code; this has not changed.

PR #818, opened 2025-08-18
feat: centralize env variable checks in config

### LINKED ISSUE #245: Better env var management
We're doing `process.env.foo ?? ""` and checking for non-empty strings everywhere that uses the env var.

We should centralize this into a config module and perform all the checks upfront in there.

bigint:
Closes #245


brody192:
I think doing this with Zod would be the better approach.

PR #819, opened 2025-08-18
fix: prevent multiple PrismaClient instances in dev

bigint:
This PR updates the Prisma client initialization to reuse a global instance in non-production environments.

- Prevents multiple PrismaClient instances being created during hot reloads
- Ensures proper cleanup with `beforeExit` hook
- Improves stability and avoids connection limit issues during development


brody192:
We are going to be removing the feedback-related code from this codebase, so this PR won't be applicable going forward. Thank you for it nonetheless.

PR #823, opened 2025-08-18
fix: add method validation and error handling for feedback endpoint

bigint:
- Added a method guard in the feedback API to return a 405 error for non-POST requests, preventing any further processing
- Introduced error handling for feedback persistence, returning a 500 status when operations fail to stop execution early


brody192:
We are going to be removing the feedback-related code from this codebase, so this PR won't be applicable going forward. Thank you for it nonetheless.

PR #828, opened 2025-08-19
feat: add sitemap generation for docs site

bigint:
Add sitemap generation for docs site and robots.txt for better SEO indexing

Here is the sitemap generated during build

https://gist.github.com/bigint/6b80944d670826fdeef3c41798891713


brody192:
I'm personally not a fan of hardcoding specific paths as they could change.

bigint:
Ahh, just found that `next-sitemap` generates for all routes automatically. Good to know for keeping the config simple.

bigint:
It's 404 on prod https://docs.railway.com/robots.txt

Maybe `postbuild` is not called

brody192:
We now call postbuild in production, all is good.

PR #836, opened 2025-08-20
refactor: component props interfaces to remove 'export' keyword for internal use

bigint:
Refactor component props interfaces to remove 'export' keyword for internal use


bigint:
Anything I need to improve on this PR?

PR #861, opened 2025-09-11
docs: update manage projects guide with updated workspace aware project fetching

stwalsh4118:
- Updates project fetch query to use the root project object instead of getting projects from the me object, since projects do not get attached to the me object anymore
- Add note to use workspace auth token

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -12,7 +12,6 @@ const nextConfig = withContentlayer({
   },
   images: {
     remotePatterns: [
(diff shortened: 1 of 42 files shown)


brody192:
Hey, could you please remove the plugins from this query? All plugins have been removed from the platform for well over a year at this point.

stwalsh4118:
> Hey, could you please remove the plugins from this query? All plugins have been removed from the platform for well over a year at this point.

Just did! Thanks for the heads up.

PR #868, opened 2025-09-19
Update template kickback docs

itsRems:
Updates the template kickback documentation following the release of Stripe Connect for cash withdrawals.


brody192:
The "Why is my country not supported?" section is good, but I think we should explicitly, somewhere on that page, perhaps inside a banner, call out that some countries are not supported, and give examples of the three largest countries that aren't supported.

I would hate for someone to go through creating a template, only to later find out they can't withdraw to cash.

PR #900, opened 2025-10-30
Include TailwindCSS & SolidQueue in Rails Guides

patrickemuller:


patrickemuller:
> There are many indent changes. Can you revert those?

I'll do my best, this was done by my Rubocop config, sorry for that.

PR #914, opened 2025-11-07
Remove legacy waived costs info from docs

kadumedim:
- removed the legacy waived seat cost messaging from the workspace reference doc
- clarify that adding or removing members carries no additional seat cost


brody192:
Thank you very much!

PR #920, opened 2025-11-18
feat: storage buckets docs v1

noahdunnagan:
Add some storage buckets documentation.

### DIFF
--- a/src/docs/guides/storage-buckets.md
+++ b/src/docs/guides/storage-buckets.md
@@ -22,7 +22,8 @@ Storage Buckets are fully S3-compatible, so you can connect with any S3-capable 
 
 In some instances when attempting to establish a connection you’ll see things like:
 
-`Invalid client configuration: A behavior major version must be set when sending a request or constructing a client.`
+`Invalid client configuration:`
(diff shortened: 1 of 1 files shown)


src/docs/guides/storage-buckets.md
```diff
@@ -0,0 +1,59 @@
...
+  });
+```
+
+### Variable References
+
+Storage Buckets support [variable reference](https://docs.railway.com/guides/variables#referencing-a-shared-variable) in two ways. The first is manually entering fields you want linked by navigating to your service’s variable tab and selecting “Add Reference”.
+
+<video src="https://res.cloudinary.com/railway/video/upload/v1763419449/VariableReferenceManual_ld79zb.mp4" controls autoplay loop muted playsinline />
+
+The second method injects pre-configured variable styles into your service. Select your bucket, open the Credentials tab, choose “Add to Service”, pick the service and style, then click “Add Variables”. Redeploy the service for the changes to take effect.
```
brody192: Took me a bit too long to realize what you were meaning by "variable styles" when first mentioned in this sentence.

brody192:
I think you should mention the limitations, such as the fact that buckets are private only.

PR #944, opened 2025-12-02
Update Docs About Presigned URLs with Storage Buckets

danieldiekmeier:
Thank you for all the new docs about storage buckets! It's very helpful, especially the new parts about presigned URLs. 

This PR only has a few tiny adjustments:

- "other's" instead of "others"
- Consistent "Use-cases" instead of sometimes "Example use-case/s"
- Mention that the service is still the redirect (if only so that nobody can Well-Actually you)


timomeh:
Thanks a lot, much appreciated!

timomeh (approved):

PR #966, opened 2025-12-16
Make better use of build cache

imax9000:
Usually the point of copying go.mod and go.sum separately is to let `docker build` use cached `go mod download` layer. Copying in all other sources before `go mod download` defeats that purpose and would trigger a full download on every build if the code has changed.


MykalMachon:
Closing this PR as it's stale and now has merge conflicts.
Feel free to re-open with merge conflicts resolved if you'd still like to contribute 👍🏻

PR #1144, opened 2026-05-08
docs(serverless): first request 502 bad gateway response caveat

uxuz:
Document potential 502 Bad Gateway response for the first request that is sent to a slept service.


brody192:
Could we add this caveat as its own independent bullet point instead?

uxuz:
Done 😃

PR #1150, opened 2026-05-12
Add ParadeDB to "Deploy an app in minutes"

philippemnoel:
Hi team. This may be a controversial PR, but I figured I would submit it anyway.

I'd like to propose updating the "Deploy an app in minutes" from Timescale to ParadeDB. Of course, I am a maintainer of ParadeDB so I am biased. But here are my reasons:

- The template linked is not a Verified template.
- ParadeDB is funneling users to Railway, while Timescale has a competing cloud offering.
- ParadeDB, like Timescale, is a Postgres extension and so targets a similar niche, but has a more broad offering (search and analytics), unlike time series, which is likely more applicable to the broad Railway community.

Please let me know if you disagree/this is out of place, and thank you for reviewing.


philippemnoel:
Lmk :)

PR #1153, opened 2026-05-15
Document PATH behavior for SSH tmux sessions

radiantjade:
## Summary

Documents that `railway ssh --session` starts through tmux and may not inherit the same PATH adjustments visible in a direct `railway ssh` shell.

This came up in Central Station thread https://station.railway.com/questions/railway-ssh-session-has-a-different-pa-b7eea7d5 where the user specifically asked why the known behavior was not documented.

## Verification

- `git diff --check`


brody192:
This contains blatantly incorrect information.

PR #1333, opened 2026-09-03
docs(guides): deploy-tested corrections for the TanStack Start refresh

codyde:
Targeted at #1331's branch so it can merge straight into that PR. Every change here comes from deploying real scaffolds to Railway today (Railpack 0.38.0):

- **The missing-start-script story split into its two real cases.** With no server runtime at all, the srvx fallback serves the `dist/` build correctly (deployed and verified, SSR responds). With `nitro()` and no `start` script, srvx targets `dist/server/server.js` while Nitro built `.output/`, so the container crashes with `ENOENT`, the domain 502s, and the deployment flips to `CRASHED` (deployed and verified). The second case is what the scaffolder's **default** `Nitro (agnostic)` deploy option produces, so the create-app section now warns about that prompt choice by name.
- **New troubleshooting entry:** `@tanstack/react-start` in `devDependencies` makes detection miss, and the app deploys as `vite static site` with 404s everywhere (deployed and verified).
- Also verified while testing, no text change needed: the healthcheck-3xx entry is correct (a 302 healthcheck path failed the deploy at the configured timeout), and the interactive scaffolder's prompt is labeled **Deploy**.

Companion to #1332, which carries the larger rewrite; happy to converge the two however you prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/content/guides/tanstack-start.md
+++ b/content/guides/tanstack-start.md
@@ -30,7 +30,9 @@ Ensure [Node](https://nodejs.org/en/download) is installed, then create a new pr
 npx @tanstack/cli@latest create
 ```
 
-Follow the prompts to choose your preferred options. When you reach the deployment options, **select `railway`**. This adds a Nitro server, a `start` script, and everything else Railway needs to build and run your app — see [Choose a production server](#choose-a-production-server) below for what it does and why it matters.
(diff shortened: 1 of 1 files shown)


codyde:
Folding these corrections into #1332 instead, which now builds on the #1331 branch directly - one PR to review.

PR #1357, opened 2026-09-17
Add tracing docs

lucacasonato:
Railway now traces requests from the edge through your services and receives OpenTelemetry spans through an on-host collector — the docs had nothing on it, and a couple of guides still said Railway doesn't do tracing at all.

- **Overview** (`/observability/tracing`): how a trace flows through edge, proxy and service, enabling tracing per project and per service, sample rate behaviour (default 100%, an incoming `traceparent` is honoured), the provided `OTEL_*` variables, the Traces page, the filter fields, retention and limits, and troubleshooting.
- **Automatic instrumentation**: the OBI switch, what it captures and what it doesn't.
- **Per-language pages**: Node.js (incl. Next.js), Deno, Python, Go, Java, Ruby, .NET, Rust and PHP — install, start command, framework notes, a custom span and a verify step each.
- **Cross-links**: the observability index and the incident, OpenTelemetry and third-party observability guides now point at built-in tracing instead of saying it doesn't exist.

Everything platform-side (variable values, traces-only receiver, sampling logic, filter fields, retention, span limits, the `x-railway-trace-id` header) is checked against `railwayapp/mono`. SDK commands are checked against the upstream OpenTelemetry docs. `pnpm build` passes.

Two things to note: the pages carry the preview note since tracing is still behind the admin-only `TRACING` flag, and the default sample rate is documented as 100% while the edge tunable in `mono` still reads 1 in 100 — that change needs to land before this does.


lucacasonato:
The edge default this was waiting on has landed: railwayapp/mono#39946 sets `hikari.tracing.sample_every` to 1, so "traces 100% of requests" is true once it rolls out.

4aca7e3 brings the pages up to date with railwayapp/mono#39943 and railwayapp/mono#39950: the two sampler variables a project with its own rate gets, and what they do for traces the service starts itself; the per-service override as the **Traced** switch in the panel or the selector under **Settings → Tracing**; automatic instrumentation picked from the two cards; and a tracing section in the variables reference. `pnpm build` still passes.

